PR #53, opened 2022-02-10
File Name Convention Integration

collinss-jpl:
This PR adds integration of file name assignment for output products which may be tailored on a per-PGE basis to meet the file naming conventions expected for that PGE. Currently, this includes a base implementation within the base PGE post-processor class, as well as a tailored implementation for the existing DSWx-HLS PGE.

The file-renaming step occurs during post-processing during staging of all output products to the output location defined by the RunConfig. The Base PGE post-processor iterates over all products created by SAS execution, and renames each file based on the file's extension. During this step, the catalog metadata and PGE/SAS log is also written to disk using the appropriate file names. 

Inheritors of the Base PGE need only provide their own implementations for their specific renaming functions, leaving the rest of the logic to the implementation in the Base PGE.


src/opera/pge/base_pge.py
```diff
@@ -247,39 +237,20 @@ def _run_sas_qa_executable(self):
...
-            msg = "Successfully created catalog metadata json file."
-            self.logger.info("pge_main", ErrorCode.CREATING_CATALOG_METADATA, msg)
-        else:
-            msg = f"SCHEMA ERROR: {met_file.get_error_msg()}"
-            self.logger.info("pge_main", ErrorCode.CREATING_CATALOG_METADATA, msg)
 
-    def _create_iso_metadata(self):
-        # TODO
-        pass
+        return MetFile(catalog_metadata)
```
JimHofman, line 251: Very nice clean factoring of the code.

src/opera/pge/base_pge.py
```diff
@@ -295,6 +266,196 @@ def _finalize_log(self):
...
+
+        The GeoTIFF filename for the Base PGE consists of:
+
+            <Core filename>_<inter_filename>.tif
+
+        Where <Core filename> is returned by PostProcessorMixin._core_filename()
+
+        Parameters
+        ----------
+        inter_filename : str, optional
```
JimHofman, line 312: Is this really option.  Won't line 323 fail if this is not defined?

src/opera/pge/dswx_pge.py
```diff
@@ -143,11 +231,8 @@ def run_postprocessor(self, **kwargs):
         print(f'Running postprocessor for {self._post_mixin_name}')
 
         self._run_sas_qa_executable()
-        self._create_catalog_metadata()
-        self._create_iso_metadata()
         self._validate_output()
         self._stage_output_files()
```
JimHofman, line 235: Again nice factoring.

JimHofman (approved):

src/opera/pge/base_pge.py
collinss-jpl: Good find, I definitely wanted the interface to allow `inter_filename` to be optional, but this function, and the version in `dsxw_hls.py` should also be checking for `None` and handling appropriately.

PR #62, opened 2022-02-24
flake8 is ok, minor improvement in code coverage

JimHofman:
All the files have passed the linting test with the exception of test_img_utils.py when an expected _import unused_ for _osgeo.gdal_ was flagged.   Wrote a unit test for __geotiff_filename_.  Only able to add a little bit to the code coverage.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,8 +2,9 @@
 
 """The setup script."""
 
-import opera
 from setuptools import find_packages, setup
+
+import opera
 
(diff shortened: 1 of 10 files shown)


collinss-jpl (approved):
@JimHofman Looks great! Looking at the current flake8/pylint reports in Jenkins, there are definitely a few findings we could annotate so they're ignored for future runs (such as the osego import finding), but we can worry about that later.

PR #101, opened 2022-05-27
98 docker stats

JimHofman:
Cleaned  up shell scripts and added a basic process_metric_data.py.


collinss-jpl (approved):
@JimHofman had to make a few tweaks, but otherwise I was able to get the branch working without issue. Thanks for your hard work, this is a great first step for this feature. Thanks also to @david-inglish for supporting Jim with the branch along the way.

PR #102, opened 2022-05-31
File rename map wildcard support

collinss-jpl:
This PR updates the file-rename-by-extension map to be a file-rename-by-pattern-map, with the map keys supporting unix-style wildcards (*) for more flexible mapping of file names to associated rename functions.


david-inglish (commented):
The changes look good.  I don't have any suggestions for improvements.

PR #103, opened 2022-05-31
Issue94 mgrs tile code to lat lon utility

david-inglish:
Initial implementation of utility to convert from MGRS tile codes to lat/lon bounding box.

### DIFF
--- a/src/opera/test/util/test_metadata_utils.py
+++ b/src/opera/test/util/test_metadata_utils.py
@@ -1,0 +1,60 @@
+#!/usr/bin/env python
+
+#
+# Copyright 2022, by the California Institute of Technology.
+# ALL RIGHTS RESERVED.
+# United States Government sponsorship acknowledged.
+# Any commercial use must be negotiated with the Office of Technology Transfer
+# at the California Institute of Technology.
+# This software may be subject to U.S. export control laws and regulations.
+# By accepting this document, the user agrees to comply with all applicable
+# U.S. export laws and regulations. User has the responsibility to obtain
+# export licenses, or other export authority as may be required, before
+# exporting such information to foreign countries or providing access to
+# foreign persons.
+#
+
+"""
+=================
+test_metadata_utils.py
+=================
+
+Unit tests for the util/metadata_utils.py module.
+
+"""
+
+import unittest
+from unittest import skipIf
+
+def mgrs_osr_gdal_are_available():
+    """
+    Helper function to check for a local installation of the Python bindings for
+    the Geospatial Data Abstraction Library (GDAL), osr and mgrs.
+    Used to skip tests that require GDAL if it's not available.
+    """
+    try:
+        from osgeo import gdal  # noqa: F401
+        from osgeo import osr  # noqa: F401
+        import mgrs  # noqa: F401
+        return True
+    except (ImportError, ModuleNotFoundError):
+        return False
+
+
+class MetadataUtilsTestCase(unittest.TestCase):
+    """Unit test Metadata Utilities"""
+
+    @skipIf(not mgrs_osr_gdal_are_available(), reason="gdal, osr, and/or mgrs is not installed on the local instance")
+    def test_get_geographic_boundaries_from_mgrs_tile(self):
+        from opera.util.metadata_utils import get_geographic_boundaries_from_mgrs_tile
+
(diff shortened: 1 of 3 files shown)


src/opera/test/util/test_metadata_utils.py
```diff
@@ -0,0 +1,60 @@
...
+    except (ImportError, ModuleNotFoundError):
+        return False
+
+
+class MetadataUtilsTestCase(unittest.TestCase):
+    """Unit test Metadata Utilities"""
+
+    @skipIf(not mgrs_osr_gdal_are_available(), reason="gdal, osr, and/or mgrs is not installed on the local instance")
+    def test_get_geographic_boundaries_from_mgrs_tile(self):
+        from opera.util.metadata_utils import get_geographic_boundaries_from_mgrs_tile
```
collinss-jpl, line 49: Please co-locate this with the rest of the imports at the top of the module

src/opera/util/metadata_utils.py
```diff
@@ -0,0 +1,95 @@
...
+
+ISO metadata utilities for use with OPERA PGEs.
+
+"""
+
+from osgeo import gdal
+from osgeo import osr
+import mgrs
+
+def get_geographic_boundaries_from_mgrs_tile(mgrs_tile_name, verbose=False):
```
collinss-jpl, line 17: A nice to have for this function would be if it could check `mgrs_tile_name` to see if it starts with a "T", and if so, remove it. The tile codes we parse from input HLS datasets usually start with a "T", but `mgrs.MGRSToUTM` does not expect it, so this function is the best place to do the input sanitation.

src/opera/test/util/test_metadata_utils.py
```diff
@@ -0,0 +1,60 @@
+#!/usr/bin/env python
+
+#
+# Copyright 2022, by the California Institute of Technology.
```
collinss-jpl, line 4: This whole comment block can be removed, as we now maintain a single copyright statement at the top of the repo

src/opera/util/metadata_utils.py
```diff
@@ -0,0 +1,95 @@
...
+    lat_max : float
+        maximum latitude of bounding box
+    lon_min : float
+        minimum longitude of bounding box
+    lon_max : float
+        maximum longitude of bounding box
+    """
+ 
+    mgrs_obj = mgrs.MGRS()
+    lower_left_utm_coordinate = mgrs_obj.MGRSToUTM(mgrs_tile_name)
```
collinss-jpl, line 38: This just occurred to me, but we should probably wrap this call in a try/except for when an unknown or otherwise invalid tile code is provided. This would make a good additional test case as well.

collinss-jpl (commented):
Branch looks good, had some minor findings and suggestions for improvements, but I'm most concerned about making sure the unit test runs from within the container. See my comment on the `skipIf` decorator line.

src/opera/test/util/test_metadata_utils.py
david-inglish: I think I need more input on this.  I moved this into the @skipIf scope because otherwise it will raise an exception in get_geographic_boundaries_from_mgrs_tile() when it tries to import osgeo if it doesn't exist.  I don't think that we want to suppress the import exception in get_geographic_boundaries_from_mgrs_tile().

One thought is that I can add try/except to the import in test_metadata_utils.py:

try:
    # If mgrs, osr, or gdal are not available for import, this will raise an exception
    from opera.util.metadata_utils import get_geographic_boundaries_from_mgrs_tile
except:
    pass

Am I missing a more elegant solution for this?

Here is the exception raised if moved out of the skipIf scope:
```
Traceback:
../../../.conda/envs/opera/lib/python3.8/importlib/__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
src/opera/test/util/test_metadata_utils.py:15: in <module>
    from opera.util.metadata_utils import get_geographic_boundaries_from_mgrs_tile
src/opera/util/metadata_utils.py:13: in <module>
    from osgeo import gdal
E   ModuleNotFoundError: No module named 'osgeo'
```

collinss-jpl (approved):
Updates all look great, thanks for the quick turnaround!

PR #105, opened 2022-06-01
ISO metadata temporal/spatial extent updates

collinss-jpl:
This PR adds the final updates to the ISO metadata for the DSWx-HLS PGE to address all feedback we received from PO.DAAC. Updates include:

- Addition of lat/lon bounding box extent to ISO metadata, leveraging the new MGRS to lat/lon utility
- Corrections to the temporal extent in the ISO metadata to specify a begin/end time range
- Replacement of the Built up cover fraction file with the Worldcover file in the ancillary input section of the ISO metadata


david-inglish (commented):
Looks good.

PR #117, opened 2022-06-14
Repo reorginaztion

collinss-jpl:
This PR reorganizes the current repo structure such that the DSWx-HLS specific code is kept within its own sub-module, along with all its ancillary files (yamale schema, jinja template, etc...). Going forward, additional PGE's will follow this convention so our Docker images can be built only with the code needed to run the corresponding PGE.


david-inglish (approved):
Looks good.

PR #127, opened 2022-06-28
CSLC-S1 PGE Stub Implementation

collinss-jpl:
Initial stub implementation for the CSLC-S1 PGE to support the SAS interface delivery.

The current stub inherits all functionality from the Base PGE. Scripts to build and test the Docker container are also included.


david-inglish (approved):
Looks good.  I don't see any issues.

PR #138, opened 2022-07-26
121 stats functionality

JimHofman:


.ci/scripts/process_metric_data.py
```diff
@@ -36,27 +41,214 @@ def remove_unwanted_lines(csv_file):
...
+    """Split into blocks read and written."""
+    r, w = block_str.split(' / ')
+    r = conv_to_gb(r)
+    w = conv_to_gb(w)
+    return r, w
+
+
+def format_out_row_misc(misc_row):
+    """
+       Return a formatted, comma separated row of docker miscellaneous data.
```
collinss-jpl, line 126: Docstring spacing a bit off here

.ci/scripts/process_metric_data.py
```diff
@@ -36,27 +41,214 @@ def remove_unwanted_lines(csv_file):
...
+    Return a formatted, comma separated row of docker stats data.
+
+    Parameters
+    ----------
+    stats_row : dictionary
+        row from data collection of docker stats
 
+    Returns : string
+    -------
+        formatted, comma separated string
```
collinss-jpl, line 165: Extra whitespace here as well

.ci/scripts/process_metric_data.py
```diff
@@ -36,27 +41,214 @@ def remove_unwanted_lines(csv_file):
...
+    csv_to_list : list
+        List that is now ready to be formatted
+
+    """
+    csv_to_list = []
+    with open(csv_file) as csv_handle:
+        reader = csv.DictReader(csv_handle)
+        for row in reader:
+            csv_to_list.append(row)
+    return csv_to_list
```
collinss-jpl, line 200: Could condense this function a bit:

```python
with open(csv_file) as csv_handle:
    return [row for row in csv.DictReader(csv_handle)]
```

.ci/scripts/run_metrics.sh
```diff
@@ -1,15 +1,13 @@
...
 
 # run_metrics()
-#
-# This script wraps a "docker run" command with metrics_collection_start() and
+# This script wraps a "docker run" command within metrics_collection_start() and
 # metrics_collection_end() function calls.
 #
 # usage:
 #     run_metrics.sh <pge> <run config_fn> (file name only) <data_dir> (full path) <image name>
+#     Eg. run_metrics  DSWX_HLS_PGE  dswx_hls.yaml  /Users/..../l30_greenland  opera/proteus:mid_may_2022
```
collinss-jpl, line 9: 👍

.ci/scripts/run_metrics.sh
```diff
@@ -20,25 +18,31 @@ PGE=$1          # pge to run in docker
...
 
-echo "Running pge '$PGE' (image $IMAGE_NAME) using run config '$RUNCONFIG'"
+echo "Running pge '{$PGE}' (image $IMAGE_NAME) using run config '$RUNCONFIG'"
 echo "Sending 'docker run' command"
 
+#docker run --rm --name "$PGE" -u $UID:$(id -g)\
+#  -v "${DATA_DIR}"/runconfig:/home/conda/runconfig:ro \
+#  -v "${DATA_DIR}"/input_dir:/home/conda/input_dir:ro \
+#  -v "${DATA_DIR}"/output_dir:/home/conda/output_dir \
+#  -i --tty opera/proteus:mid_may_2022 sh -ci "python3 proteus-0.1/bin/dswx_hls.py runconfig/dswx_hls.yaml --log output_dir/test_log.log"
```
collinss-jpl, line 37: Can this be removed? If not we should add a TODO to indicate why we're keeping it around commented out

.ci/scripts/run_metrics.sh
```diff
@@ -20,25 +18,31 @@ PGE=$1          # pge to run in docker
...
 docker run --rm --name "${container_name}" -u $UID:$(id -g)\
-  -v ${DATA_DIR}/runconfig:/home/conda/runconfig:ro \
-  -v ${DATA_DIR}/input_dir:/home/conda/input_dir:ro \
-  -v ${DATA_DIR}/output_dir:/home/conda/output_dir \
-  -i --tty ${IMAGE_NAME} \
-  --file /home/conda/runconfig/${RUNCONFIG}
+  -v "${DATA_DIR}"/runconfig:/home/conda/runconfig:ro \
+  -v "${DATA_DIR}"/input_dir:/home/conda/input_dir:ro \
+  -v "${DATA_DIR}"/output_dir:/home/conda/output_dir \
+  -i --tty opera/proteus:mid_may_2022 sh -ci "python3 proteus-0.1/bin/dswx_hls.py runconfig/dswx_hls.yaml --log output_dir/test_log.log"
```
collinss-jpl, line 43: This line looks like it still needs to be parameterized with `${IMAGE_NAME}`, `${RUNCONFIG}`, etc..

It also looks like its still using the calling convention of the SAS-only image (the `sh -ci python3 proteus-0.1/bin/dswx_hls.py...` part), where it should be set up to call our PGE images (`--file ${RUNCONFIG}`)

.ci/scripts/util.sh
```diff
@@ -2,65 +2,129 @@
...
-metrics_collection_start ()
+# Start the metrics collection of both docker stats and the miscellaneous os statistics.
+# Parameters:
+#     container_info:  the name of the docker file from which to gather statistics
+#     sample_time:  The time between sampling of the statistics.
+metrics_collection_start()
 {
-    # container name should always contain pge_docker_tag, PGE, and runconfig (at a minimum)
-    container_name=$1
+    local pge=$1
```
collinss-jpl, line 11: Missing documentation for this new first argument in the description above

.ci/scripts/util.sh
```diff
@@ -2,65 +2,129 @@
...
+    # Split seconds argument into docker name and tag
+    local cont_name="$(echo "$2" | cut -d':' -f1)"
+    local container_tag="$(echo "$2" | cut -d':' -f2)"
+    # cont_info should always contain pge_docker_tag, PGE, and runconfig (at a minimum)
+    # container_info="${pge}_${cont_name}_${container_tag}"
 
+    local container_info="$pge"
+
+    # If no sample_time value is passed - default to a value of 1
+    if [[ -z "$2" ]]
```
collinss-jpl, line 21: Should this be `if [[ -z "$3" ]]` now that there's an additional input argument?

.ci/scripts/util.sh
```diff
@@ -2,65 +2,129 @@
...
+        local sample_time=1
+    else
+        local sample_time=$3
+    fi
+
+    echo "Using sample time of: $sample_time"
     # Initialize output files and statistics format
-    metrics_stats="${container_name}_metrics_stats.csv"
+    metrics_stats="${container_info}_metrics_stats.csv"
+    echo "local file name from assignment at line 30: $metrics_stats"
```
collinss-jpl, line 31: Is this a leftover debug statement? Seems strange to have the hardcoded reference to line 30 here....

.ci/scripts/util.sh
```diff
@@ -2,65 +2,129 @@
...
+        { while true; do dus=$(eval "$block_space_cmd"); swu=$(eval "$swap_space_cmd"); ths=$(eval "$sys_threads_cmd"); \
+        lll=$(eval "$lll_cmd"); echo "$(metrics_seconds), $dus, $swu, $ths, $lll" >> "${metrics_misc}"; \
+        sleep "$sample_time"$; done } & \
+        echo "$!" >> "${container_info}_metrics_misc_bg_pid.txt"
+    fi
+
 }
 
-metrics_collection_end ()
+metrics_collection_end()
```
collinss-jpl, line 96: Would be good to have a comment that describes the input arguments, similar to what is in `metrics_collection_start()`

src/opera/scripts/pge_main.py
```diff
@@ -177,7 +177,9 @@ def pge_main():
     run_config_filename = os.path.abspath(args.file)
 
     if not os.path.exists(run_config_filename):
-        raise FileNotFoundError(f"Could not find config file: {run_config_filename}")
+        msg = f"Could not find config file: {run_config_filename}: argument entered {run_config_filename}"
```
collinss-jpl, line 180: Not sure what's going on here, more leftover debug code? Can these changes be reverted?

collinss-jpl (changes requested):
Updates look good, but I noticed a couple things that need to be addressed before merge. See my findings below.

.ci/scripts/process_metric_data.py
```diff
@@ -36,27 +41,214 @@ def remove_unwanted_lines(csv_file):
...
+    # convert = {'SECONDS': int(), 'Name': str(), 'PIDs': int(), 'CPU': float(), 'Memory': float(), 'MemoryP': float(),
+    #            'NetSend': float(), 'NetRecv': float(), 'DiskRead': float(), 'DiskWrite': float()}
+    with open(docker_report_file, 'w') as out_file:
+        out_file.write(f"{docker_columns}\n")
+        for stats_row in stats_list:
+            row = format_out_row_docker(stats_row)
+            out_file.write(f"{row}\n")
+
+    # Write out the miscellaneous file
+    # Todo - divide out mac OS runs and linux runs - Swap will work on linux
```
david-inglish, line 239: See my comment in the shell script about keeping a column for swap space even if it is not used.

.ci/scripts/process_metric_data.py
JimHofman: fixed

.ci/scripts/process_metric_data.py
JimHofman: fixed

.ci/scripts/process_metric_data.py
JimHofman: Cool application of a list comprehension.   I'll look for opportunities to use.

.ci/scripts/run_metrics.sh
JimHofman: Thanks, I meant to delete that.

.ci/scripts/util.sh
JimHofman: Thanks.  Good catch.

.ci/scripts/util.sh
JimHofman: Yes, however, this will all go away when I store intermediate data in memory rather than writing temp files.

.ci/scripts/util.sh
JimHofman: Done

src/opera/scripts/pge_main.py
JimHofman: Done

.ci/scripts/util.sh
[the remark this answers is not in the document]
JimHofman: Done

.ci/scripts/util.sh
```diff
@@ -2,65 +2,129 @@
...
-    # TODO For the time being (may ultimately remove) take out the sleep ("$(metrics_seconds),$ds" >> ${metrics_stats}; sleep .1; done })
-    { while true; do ds=$(docker stats --no-stream --format "${stat_format}" "${container_name}" 2>/dev/null); echo "$(metrics_seconds)","$ds" >> "${metrics_stats}"; done } &
-    echo "$!" > "${container_name}_metrics_stats_bg_pid.txt"
+    { while true; do ds=$(docker stats --no-stream --format "${stat_format}" "${pge}" 2>/dev/null); \
+    echo "$(metrics_seconds)","$ds" >> "${metrics_stats}"; sleep "$sample_time"; done } & \
+    echo "$!" > "${container_info}_metrics_stats_bg_pid.txt"
+
+    # Miscellaneous Statistics
+
+    # test for MAC operating system for OS statistics
```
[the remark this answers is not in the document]
JimHofman: Done

.ci/scripts/process_metric_data.py
JimHofman: Done

.ci/scripts/util.sh
david-inglish: I would like to continue to see these files kept for a couple of reasons.  One reason is that it is the original record of the metrics which can be imported into excel for quick analysis, or can be re-processed by a script at a later date to create other types of reports.  An example of post-processing might be to create a graph showing a parallel processing PGE's run-time versus vCPU quantity; you might want to reference the raw metrics data from many docker container executions (in Excel or a script).
Another reason is if there is a program interruption, a file will capture at least some of the data.  I had this happen with a multi-hour job where metrics were collected, but the terminal connection to the AWS instance was lost before the process_metrics_data.py script was run.  I was able to get the information I needed from the partial file.

collinss-jpl:
Hi @JimHofman, changes all look good, but it looks like this one finding got overlooked: https://github.com/nasa/opera-sds-pge/pull/138#discussion_r931468722 (maybe because github folded it into a "hidden" conversation). Unfortunately, this one will need to be addressed before the branch is merged, but it should be a pretty simple fix.

collinss-jpl (approved):

.ci/scripts/util.sh
JimHofman: I no longer delete those two files

PR #145, opened 2022-08-08
CSLC-S1 Filename Convention Application

collinss-jpl:
This branch adds an implementation for file name application to the CSLC-S1 PGE.

The information required to fill in the components of the final filename are obtained from the JSON metadata file produced by the SAS.

This branch also modifies the `get_output_product-filenames()` method of the `RunConfig` class to support finding of output products within an arbitrary number of sub-directories.


david-inglish (approved):
I don't see any issues with the changes.  There's a very minor typo in src/opera/pge/cslc_s1/cslc_s1_pge.py at line 86, "needed" should probably be "needing".

PR #150, opened 2022-08-18
DSWx-HLS Filename Convention and Int Test Updates

collinss-jpl:
This PR addresses the recent change to the output product filename format for DSWx-HLS products where some fields were shuffled around.

I wanted to try this branch out with the Integration Test pipeline, so I also ended up make some modifications to allow the pipeline to support multiple PGE's, in expectation of the CSLC-S1 test script. I also updated the test assets for DSWx to the versions received from ADT for the CalVal delivery.


david-inglish (approved):
Looks good

PR #163, opened 2022-08-24
Return to starting directory in tearDown(), so temporary directory is cleaned up appropriately.

daiangan:
The solution was to replace __os.chdir(self.test_dir)__ by __os.chdir(self.starting_dir)__ on __tearDown()__ method to correctly return to starting directory, so temporary directory is cleaned up appropriately without leaving a __test_logger_xxxxxtemp__  directory within whatever working directory the test was started in.

### DIFF
--- a/src/opera/test/util/test_logger.py
+++ b/src/opera/test/util/test_logger.py
@@ -74,7 +74,7 @@ class LoggerTestCase(unittest.TestCase):
 
     def tearDown(self) -> None:
         """Return to starting directory"""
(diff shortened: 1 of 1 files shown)


collinss-jpl (approved):
Works for me! Thanks again for jumping on this @daiangan, had been driving me crazy! 🙇

daiangan:
Thanks for the merge @collinss-jpl 
Running the tests I noticed that there are a lot of styles that can be improved to pass flake8.
Can I go ahead with this?

PR #164, opened 2022-08-24
Improve Python styles to cleanup and pass flake8.

daiangan:
I made some changes on imports to improve Python styles, cleanup and pass flake8 with the command:

```text
flake8 --config .flake8 --application-import-names src
```

### DIFF
--- a/.ci/scripts/process_metric_data.py
+++ b/.ci/scripts/process_metric_data.py
@@ -1,11 +1,5 @@
 #!/usr/bin/env python3
 
-import csv
-import os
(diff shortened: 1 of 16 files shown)


collinss-jpl:
Hi @daiangan, thanks for putting this branch together. 

On OPERA we actually maintain our own flake8 configuration that is tied into code quality checks we do as part of our CI/CD, so our findings may be different than the ones you've addressed here. So things don't get too out of sync, I'm going to close this PR for now.

PR #165, opened 2022-08-24
HLS Sensing Time parsing fix

collinss-jpl:
This branch fixes an issue where multiple timestamps concatenated with a plus sign assigned to the `SENSING_TIME` field of DSWx-HLS metadata were propagated to the ISO metadata, resulting in validation errors.

Logic has been added to the method that collects DSWx product metadata to detect sensing time fields containing a "+", and simply parse out the first timestamp for use as the begin/end value.


david-inglish (approved):
I don't see any issues with the changes.

PR #170, opened 2022-09-15
RTC-S1 Stub Implementation

collinss-jpl:
Initial addition of the stub implementation for the RTC-S1 PGE. 

Note that due to the specifics of how this PGE/SAS works, I also had to add a partial implementation of the file-naming conventions to ensure the PGE does not clobber the multiple per-burst outputs from the SAS.

I also took the liberty of refactoring some common code out of the build/test CI scripts into util.sh to keep things maintainable as we add more and more of these scrips.

### DIFF
--- a/.ci/scripts/build_cslc_s1.sh
+++ b/.ci/scripts/build_cslc_s1.sh
@@ -1,39 +1,13 @@
 #!/bin/bash
 set -e
 
+# Source the build script utility functions
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
+
+. "${SCRIPT_DIR}"/util.sh
+
 # Parse args
-while [[ $# -gt 0 ]]; do
-  case $1 in
-    -h|--help)
-      echo "Usage: build_cslc_s1.sh [-h|--help] [-s|--sas-image <image name>] [-t|--tag <tag>] [-w|--workspace <path>]"
-      exit 0
-      ;;
-    -s|--sas-image)
-      SAS_IMAGE=$2
-      shift
-      shift
-      ;;
-    -t|--tag)
-      TAG=$2
-      shift
-      shift
-      ;;
-    -w|--workspace)
-      WORKSPACE=$2
-      shift
-      shift
-      ;;
-    -*|--*)
-      echo "Unknown arguments $1 $2, ignoring..."
-      shift
-      shift
-      ;;
-    *)
-      echo "Unknown argument $1, ignoring..."
-      shift
-      ;;
-  esac
-done
+parse_build_args "$@"
 
 echo '
 =====================================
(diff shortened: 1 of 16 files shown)


src/opera/test/pge/rtc_s1/test_rtc_s1_pge.py
```diff
@@ -0,0 +1,132 @@
...
+
+        os.system(f"touch {join(input_dir, 'ORBIT.EOF')}")
+
+        os.system(f"touch {join(input_dir, 'dem.tif')}")
+
+        os.chdir(self.working_dir.name)
+
+    def tearDown(self) -> None:
+        """Return to test directory"""
+        os.chdir(self.test_dir)
```
david-inglish, line 69: This looks like it is always "None".  Is this desired behavior?

src/opera/test/util/test_logger.py
```diff
@@ -74,7 +76,7 @@ def setUp(self) -> None:
 
     def tearDown(self) -> None:
         """Return to starting directory"""
```
david-inglish, line 78: This comment may need an update?

david-inglish (approved):
I've made some notes where I had questions, but they don't seem to be major issues being in the unit test code.

.ci/scripts/build_cslc_s1.sh
```diff
@@ -1,39 +1,13 @@
...
-      shift
-      shift
-      ;;
-    *)
-      echo "Unknown argument $1, ignoring..."
-      shift
-      ;;
-  esac
-done
+parse_build_args "$@"
```
JimHofman, line 10: Not sure what magic is going on here, but it sure saves some lines of code.

.ci/scripts/util.sh
```diff
@@ -2,6 +2,119 @@
 
 SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
 
+parse_build_args()
```
JimHofman, line 5: I see good idea to factor this out.

src/opera/test/data/test_rtc_s1_config.yaml
```diff
@@ -0,0 +1,174 @@
...
+            DynamicAncillaryFilesGroup:
+                AncillaryFileMap:
+                    dem_file: rtc_s1_test/input_dir/dem.tif
+
+            ProductPathGroup:
+                OutputProductPath: rtc_s1_test/output_dir
+                ScratchPath: rtc_s1_test/scratch_dir
+
+            PrimaryExecutable:
+                ProductIdentifier: RTC_S1
```
JimHofman, line 23: Will we be adding ProductVersion here as in the other PGEs?

src/opera/test/data/test_rtc_s1_config.yaml
```diff
@@ -0,0 +1,174 @@
...
+                ProgramPath: mkdir
+                ProgramOptions:
+                    - '-p rtc_s1_test/output_dir/t069_147170_iw1/;'
+                    - '/bin/echo hello world > rtc_s1_test/output_dir/t069_147170_iw1/rtc_product.nc;'
+                    - '/bin/echo RTC-S1 invoked with RunConfig'
+                ErrorCodeBase: 300000
+                SchemaPath: pge/rtc_s1/schema/rtc_s1_sas_schema.yaml
+                IsoTemplatePath:
```
JimHofman, line 31: Similiarly, will we have CompositeReleaseID (not sure if we want to add it before ticket 149 is merged but we should remember these additions if they apply.

JimHofman (approved):
We should remember that we will probably be added the changes for Product Version and Composite Release ID as described in ticket 149.

.ci/scripts/build_cslc_s1.sh
collinss-jpl: `"$@"` is bash shorthand for "array of all the command line variables", which includes the name of the script as the first variable (at $0). Basically this is just forwarding all the command line options to the parsing function which has been refactored into util.sh.

src/opera/test/data/test_rtc_s1_config.yaml
collinss-jpl: Yes, I will likely merge your branch first, then update this branch to incorporate the new fields

src/opera/test/pge/rtc_s1/test_rtc_s1_pge.py
collinss-jpl: `test_dir` is initialized to `None` at line 32, but `setUpClass` (which is called before any calls to `setUp` or `tearDown`) sets it to the location of the test directory in our repository.

Issue #157, opened 2022-08-23
HLS Product Type Validation


collinss-jpl:
Recent PCM system testing has discovered an issue where HLS L30 datasets corresponding to Landsat-7 (as opposed to Landsat-8/9) data were processed mistakenly due to incorrect labeling by the data provider. 

There are plans to improve the PCM query logic to ensure Landsat 7 data is filtered out, but the DSWx-HLS PGE should still be able to detect when it's been provided unexpected HLS data and log an exception.

The goal of this ticket is to enhance the DSWx-HLS pre-processor to examine the metadata from the input HLS S30/L30 tif files to ensure they came from an expected platform (currently Landsat 8/9, or Sentinel 2 A/B).

For L30 HLS data, this can be determined from the `LANDSAT_PRODUCT_ID` field:
```python
>>> from osgeo import gdal
>>> gdal.Open('HLS.L30.T22VEQ.2021248T143156.v2.0.B01.tif').GetMetadata()
{... 'LANDSAT_PRODUCT_ID': 'LC08_L1TP_004016_20210905_20210905_02_RT; LC08_L1TP_004017_20210905_20210905_02_RT',  ...}
```

If the ID (or ID's in this case) start with `LC07`, the data corresponds to Landsat-7 and we would want to flag this. If it starts with `LC08` or `LC09`, then the data should be considered valid.

Likewise for S30, the relevant field seems to be `PRODUCT_URL`:

```python
>>> gdal.Open('HLS.S30.T15SXR.2021250T163901.v2.0.B01.tif').GetMetadata()
{... 'PRODUCT_URI': 'S2A_MSIL1C_20210907T163901_N0301_R126_T15SXR_20210907T202434.SAFE', ...}
```

Where in this case, we would want to validate that the value starts with either `S2A` or `S2B` for it to be considered valid.

PR #174, opened 2022-09-21
157 DSWx HLS product type validation

david-inglish:
These code changes add checking for Landsat-7 and non-Sentinel A/B datasets as input files.  Exceptions will be raised if these types of datasets are encountered by the PGE.

Issue description:
https://github.com/nasa/opera-sds-pge/issues/157

### DIFF
--- a/src/opera/pge/dswx_hls/dswx_hls_pge.py
+++ b/src/opera/pge/dswx_hls/dswx_hls_pge.py
@@ -94,8 +94,7 @@ class DSWxHLSPreProcessorMixin(PreProcessorMixin):
         for input_file in self.runconfig.input_files:
             input_file_path = abspath(input_file)
             if isdir(input_file_path):
-                tifs_in_dir = glob.glob(join(input_file_path, '*.tif*'))
-                list_of_input_tifs.append(tifs_in_dir)
+                list_of_input_tifs = glob.glob(join(input_file_path, '*.tif*'))
             else:
                 list_of_input_tifs.append(input_file_path)
 
--- a/src/opera/pge/dswx_hls/dswx_hls_pge.py
+++ b/src/opera/pge/dswx_hls/dswx_hls_pge.py
@@ -63,7 +63,7 @@ class DSWxHLSPreProcessorMixin(PreProcessorMixin):
 
                 self.logger.critical(self.name, ErrorCode.INPUT_NOT_FOUND, error_msg)
             elif isdir(input_file_path):
-                list_of_input_tifs = glob.glob(join(input_file_path, '*.tif'))
+                list_of_input_tifs = glob.glob(join(input_file_path, '*.tif*'))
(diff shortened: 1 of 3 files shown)


src/opera/pge/dswx_hls/dswx_hls_pge.py
```diff
@@ -74,6 +74,50 @@ def _validate_inputs(self):
...
+            f'Scanning DSWx input datasets for invalid platforms.'
+        )
+
+        # Get a list of input files to check for invalid platform metadata
+        for input_file in self.runconfig.input_files:
+            input_file_path = abspath(input_file)
+            if isdir(input_file_path):
+                list_of_input_tifs = glob.glob(join(input_file_path, '*.tif'))
+            else:
+                list_of_input_tifs = [input_file_path]
```
collinss-jpl, line 98: I think this case needs to be appending `input_file_path` to `list_of_input_tifs`. This is because inputs could be specified in one of two ways: as a directory containing the 7 or so input HLS bands (covered by your `isdir` case), and as separately enumerated paths to each input HLS band. As-is, this would overwrite `list_of_input_tifs` for each `input_file`.

src/opera/test/pge/dswx_hls/test_dswx_hls_pge.py
```diff
@@ -369,6 +388,34 @@ def test_dswx_product_metadata_collection(self):
...
+        """Test exceptions raised for Landsat 7 input data"""
+
+        # Create a filename that will trigger a metadata check
+        input_file = os.path.join(self.input_dir, "HLS.L30.T22VEQ.2021248T143156.v2.0.VAA.tif")
+        os.system(f"touch {input_file}")
+
+        runconfig_path = join(self.data_dir, 'test_dswx_hls_config.yaml')
+
+        pge = DSWxHLSExecutor(pge_name="DSWxPgeTest", runconfig_path=runconfig_path)
+        with self.assertRaises(Exception) as context:
```
collinss-jpl, line 402: Ensuring we get an Exception for these test cases is definitely what we want, but it also carries the risk of causing the test to pass if code throws any kind of unrelated Exception, essentially masking the original issue. The workaround I've tried to do for this is to add another check on the contents of the `logger` to ensure we got the expected error message (or at least an easy-to-search-for fragment). That way, we can be sure the test actually triggered the error we expected.

Lines [167-181 ](https://github.com/nasa/opera-sds-pge/blob/develop-R2/src/opera/test/pge/dswx_hls/test_dswx_hls_pge.py#L167) here are a good example you can copy pretty much wholesale, and just add the log fragment to search for.

JimHofman (commented):
Nothing to add that Scott did not already say.  Good work David.

collinss-jpl (approved):
Updates look great, this branch is good to go

PR #186, opened 2022-10-06
183 input validation

JimHofman:
Took validation code from cslc_s1 CslcS1PreProcessorMixin.
Parameterized and wrote the code to /opera/util/input_validation.py; then added the call back into CslcS1PreProcessorMixin.
Added call: validate_inputs(self.runconfig, self.logger, self.name) to run_preprocessor() in RtcS1PreProcessorMixin

Updated units for cslc-s1 and rtc-s1.  Ended up with 93% coverage for validate_inputs.py

### DIFF
--- a/src/opera/pge/rtc_s1/rtc_s1_pge.py
+++ b/src/opera/pge/rtc_s1/rtc_s1_pge.py
@@ -16,6 +16,7 @@ from pathlib import Path
 from opera.pge.base.base_pge import PgeExecutor
 from opera.pge.base.base_pge import PostProcessorMixin
 from opera.pge.base.base_pge import PreProcessorMixin
+from opera.util.input_validation import validate_slc_s1_inputs
 from opera.util.time import get_time_for_filename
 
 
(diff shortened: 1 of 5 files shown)


src/opera/pge/rtc_s1/rtc_s1_pge.py
```diff
@@ -50,7 +51,7 @@ def run_preprocessor(self, **kwargs):
         super().run_preprocessor(**kwargs)
 
         # TODO: call input validation routine here
-
+        validate_inputs(self.runconfig, self.logger, self.name)
```
collinss-jpl, line 54: Nice! Feel free to remove the TODO at line 52 since this addresses it

src/opera/util/input_validation.py
```diff
@@ -0,0 +1,93 @@
...
+        logger.critical(name, ErrorCode.INPUT_NOT_FOUND, error_msg)
+
+    ext = splitext(input_object)[-1]
+
+    if ext not in valid_extensions:
+        error_msg = f"Input file {input_object} does not have an expected file extension."
+        logger.critical(name, ErrorCode.INVALID_INPUT, error_msg)
+
+
+def validate_inputs(runconfig, logger, name):
```
collinss-jpl, line 46: Can we give this a slightly more specific name? Since the function performs the input validation for the SLC-S1 based PGE's, maybe something like `validate_slc_s1_inputs`?

collinss-jpl (approved):
Looks great, I'd just like to see one of the validation functions renamed so its clear which PGE's it should be used with, otherwise this is ready to go!

PR #187, opened 2022-10-13
176 Integration shell script refactoring initial implementation

david-inglish:
The changes in this pull request have moved common functionality into a new bash function library named test_int_util.sh to reduce the creation and maintenance effort of the integration test scripts.

### DIFF
--- a/.ci/scripts/test_int_util.sh
+++ b/.ci/scripts/test_int_util.sh
@@ -1,0 +1,142 @@
+#!/bin/bash
+#
+# test_int_util.sh
+#
+# Library of common shell routines for PGE integration testing.
+#
+set -e
+umask 002
+
+[ -z "${WORKSPACE}" ] && WORKSPACE=$(realpath $(dirname $(realpath $0))/../..)
+
+test_int_parse_args()
+{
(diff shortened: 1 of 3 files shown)


.ci/scripts/test_int_util.sh
```diff
@@ -0,0 +1,143 @@
...
+#
+# test_int_util.sh
+#
+# Library of common shell routines for PGE integration testing.
+#
+set -e
+umask 002
+
+[ -z "${WORKSPACE}" ] && WORKSPACE=$(realpath $(dirname $(realpath $0))/../..)
+[ -z "${PGE_TAG}" ] && PGE_TAG="${USER}-dev"
```
collinss-jpl, line 11: I would like to see these four lines which are outside of any function kept within their respective scripts. As-is they become side-effects of sourcing `test_int_util.sh`, which might become confusing down the line.

collinss-jpl (commented):
Refactor looks great, love seeing both scripts reduced by about 100 lines each. I only had one comment regarding avoid side-effects from just sourcing the new utility module, otherwise this is ready to go.

Issue #173, opened 2022-09-20
Rework integration test scripts to use /data/tmp for temp files


collinss-jpl:
In response to the recent issue with disk space on our CI machine, our SA provided the following:

    Hi Scott, 

    I increased /data to 300GB and added a /data/tmp directory owned by jenkins.

    Please perform any temporary work or downloads in /data/tmp rather than /tmp

This ticket is to perform the necessary changes in our existing int test scripts to use /data/tmp by default as the staging area for any S3 downloads, temporary files, etc...

riverma:
Is `/data/tmp` cleaned automatically at a set interval? What's to stop it from filling up the disk if left unchecked?

collinss-jpl:
> Is `/data/tmp` cleaned automatically at a set interval? What's to stop it from filling up the disk if left unchecked?

As far as I know there is no automatic cleanup in place. If it fills up I imagine we need to contact an SA to clear up space.

collinss-jpl:
Addressed in #187

PR #190, opened 2022-10-19
142 int stats

JimHofman:
## Description
- Correct directories for local_dir in integration test scripts


collinss-jpl:
The fix to address #173 ended up getting addressed in PR #187, closing this PR as the updates are now redundant

Issue #192, opened 2022-10-27
Use hyphen in product shortname for all file name conventions


collinss-jpl:
In all our current file naming conventions for each PGE, each file name component is joined via underscore. However, the Project convention is to always use a hyphen to join the two parts of the product short name.

For example, the output product file name `OPERA_L3_DSWx_HLS_T15SXR_20210907T163901Z_20221026T213115Z_S2A_30_v1.0_B01_WTR` is incorrect, and should be:
`OPERA_L3_DSWx-HLS_T15SXR_20210907T163901Z_20221026T213115Z_S2A_30_v1.0_B01_WTR`

Similarly for use of the CSLC-S1 and RTC-S1 product shortnames.

This ticket is to make the updates in the file naming convention functions for each PGE to ensure the "name" and "source" fields (which comprise the product shortname) are joined via hyphen ("-") and not underscore ("_"), and to update any unit test which break from the change. The file name pattern definition for each PGE should be defined within the `_core_filename` function within each PGE's post-processor mixin class.

PR #196, opened 2022-11-15
for issue #192 added hyphen

ramvikrams:
## Description
- changed files for hyphen instead of underscore and changed the respective tests for it

## Affected Issues
- #192

### DIFF
--- a/src/opera/pge/base/base_pge.py
+++ b/src/opera/pge/base/base_pge.py
@@ -346,7 +346,7 @@ class PostProcessorMixin:
         """
         time_tag = get_time_for_filename(self.production_datetime)
 
-        return f"{self.PROJECT}_{self.LEVEL}_{self.NAME}_{time_tag}"
+        return f"{self.PROJECT}_{self.LEVEL}_{self.NAME}-{time_tag}"
 
     def _geotiff_filename(self, inter_filename):
         """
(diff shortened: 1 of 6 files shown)


src/opera/pge/base/base_pge.py
```diff
@@ -346,7 +346,7 @@ def _core_filename(self, inter_filename=None):  # pylint: disable=unused-argumen
         """
         time_tag = get_time_for_filename(self.production_datetime)
 
-        return f"{self.PROJECT}_{self.LEVEL}_{self.NAME}_{time_tag}"
+        return f"{self.PROJECT}_{self.LEVEL}_{self.NAME}-{time_tag}"
```
collinss-jpl, line 349: Hi @ramvikrams, thanks for putting this PR together. My only note is that the underscore to hyphen change does not need to be applied to the placeholder filename we apply in `base_pge.py`. If you could revert this change I would be happy to merge the rest of this branch.

collinss-jpl (changes requested):

src/opera/pge/base/base_pge.py
ramvikrams: Done

collinss-jpl (approved):
Looks good, thanks again @ramvikrams for pitching in!

PR #210, opened 2022-12-15
209 tiff to tif

JimHofman:
## Description
- Changed ".tiff" to ".tif" in dswx_hls_pge.py and in test_dswx_hls_pge.py

## Testing
-  minor change to test_dswx_hls_pge.py

## Question
- in the future should we consider changing methods and variables with 'geotiff' in their names to 'geotif', to reinforce the new restrictions on the naming convention?


collinss-jpl (approved):
Thanks again for knocking this out @JimHofman!

Issue #208, opened 2022-12-09
Develop standalone version of dswx_compare.py


collinss-jpl:
In order to address some issues with the current version of dswx_compare.py (which is bundled with Proteus images), I think it will be prudent to develop/maintain a standalone version of the script for use with our integration test pipeline. This will allow us to address issues particular to our int test without ADT explicitly in the loop. Any fixes/improvements we make to our version can be communicated back to ADT, which they can choose to re-implement as they see fit.

Development of our version of the script will require the following:

1. Refactoring of the following functions into a single script/module for better portability:
-    [dswx_compare.py](https://github.com/nasa/PROTEUS/blob/main/bin/dswx_compare.py)
- [proteus.dswx_hls.compare_dswx_hls_products](https://github.com/nasa/PROTEUS/blob/36508ac0675bc81e8ab2c8ad2cbb21a55822760f/src/proteus/dswx_hls.py#L568)
- [proteus.dswx_hls._get_prefix_str](https://github.com/nasa/PROTEUS/blob/36508ac0675bc81e8ab2c8ad2cbb21a55822760f/src/proteus/dswx_hls.py#L563)
- [proteus.dswx_hls._print_first_value_diff](https://github.com/nasa/PROTEUS/blob/36508ac0675bc81e8ab2c8ad2cbb21a55822760f/src/proteus/dswx_hls.py#L690)
These functions only seem to only have dependencies on numpy and gdal, so extracting them should be straightforward.

2. Address comparison issues raised by Steven Lewis. Steven had a look at the comparison routine several months ago, and discovered a number of issues with the current logic, which are quoted below:

> The comparison code is finding a difference, but printing the wrong values to the screen.  It’s also only finding and reporting the first difference.

 > When comparing the images, a count of pixels differing would be more useful, along with some statistics for the differences (e.g. at least min and max difference (especially if the code uses fill values), and maybe some other moments can be useful).

> Also, I recommend using numpy.isclose(), instead of numpy.allclose(), to avoid needing to do two comparisons.  In the case of dswx_compare.py, numpy.allclose() is used to detect the error, and abs(image_1[i, j] - image_2[i, j]) > 1e-6 is used to find the value to print (which definitely explains why floating point errors are  reported incorrectly).

These issues were communicated to ADT, but it seems they have not had the bandwidth the address them.

[...]

PR #213, opened 2022-12-29
compare_dswx_hls_products and dependencies gathered into new standalo…

david-inglish:
…ne script

## Description
Develop a PGE version of the compare_dswx_hls_products script to enable customization for PGE purposes.

## Affected Issues
#208 

## Testing
Tested with output files from different releases

### DIFF
--- a/.ci/scripts/dswx_compare_opera_pge.py
+++ b/.ci/scripts/dswx_compare_opera_pge.py
@@ -41,54 +41,55 @@ def compare_dswx_hls_products(file_1, file_2):
     if not flag_same_nbands:
         print(prefix + f'Input 1 has {nbands_1} bands and input 2'
               f' has {nbands_2} bands')
-        return False
+    else:
+        # compare each band
+        print('Comparing DSWx bands...')
+        for b in range(1, nbands_1 + 1):
+            gdal_band_1 = layer_gdal_dataset_1.GetRasterBand(b)
+            gdal_band_2 = layer_gdal_dataset_2.GetRasterBand(b)
+            image_1 = gdal_band_1.ReadAsArray()
+            image_2 = gdal_band_2.ReadAsArray()
+            is_close_result = np.isclose(image_1, image_2,
+                                         atol=COMPARE_DSWX_HLS_PRODUCTS_ERROR_TOLERANCE_ATOL,
+                                         rtol=COMPARE_DSWX_HLS_PRODUCTS_ERROR_TOLERANCE_RTOL)
+            num_image_differences = np.sum(~is_close_result)
+            if num_image_differences > 0:
+                flag_bands_are_equal = False
+                '''
+                # Find the min and max differences using the numpy.isclose function
+                # 0 <= (atol + rtol * absolute(b)) - absolute(a-b)
+                a1 = image_1[~is_close_result]
+                b2 = image_2[~is_close_result]
+                atol = COMPARE_DSWX_HLS_PRODUCTS_ERROR_TOLERANCE_ATOL
+                rtol = COMPARE_DSWX_HLS_PRODUCTS_ERROR_TOLERANCE_RTOL
+                is_close_differences = abs((atol + rtol * abs(b2)) - abs(a1 - b2))
+                '''
+                is_close_differences = abs( image_1[~is_close_result] - image_2[~is_close_result])
+                max_difference_band = np.max(is_close_differences)
+                min_difference_band = np.min(is_close_differences)
 
-    # compare array values
-    print('Comparing DSWx bands...')
-    total_image_differences = 0
-    min_difference = 0
-    max_difference = 0
-    for b in range(1, nbands_1 + 1):
-        gdal_band_1 = layer_gdal_dataset_1.GetRasterBand(b)
-        gdal_band_2 = layer_gdal_dataset_2.GetRasterBand(b)
-        image_1 = gdal_band_1.ReadAsArray()
-        image_2 = gdal_band_2.ReadAsArray()
-        is_close_result = np.isclose(image_1, image_2,
-                                     atol=COMPARE_DSWX_HLS_PRODUCTS_ERROR_TOLERANCE_ATOL,
-                                     rtol=COMPARE_DSWX_HLS_PRODUCTS_ERROR_TOLERANCE_RTOL)
-        num_image_differences = np.sum(~is_close_result)
-        if num_image_differences > 0:
-            flag_bands_are_equal = False
-            # Find the min and max differences using the numpy.isclose function
-            # 0 <= (atol + rtol * absolute(b)) - absolute(a-b)
-            a1 = image_1[~is_close_result]
-            b2 = image_2[~is_close_result]
-            atol = COMPARE_DSWX_HLS_PRODUCTS_ERROR_TOLERANCE_ATOL
-            rtol = COMPARE_DSWX_HLS_PRODUCTS_ERROR_TOLERANCE_RTOL
-            is_close_differences = abs((atol + rtol * abs(b2)) - abs(a1 - b2))
-            max_difference_band = np.max(is_close_differences)
-            if max_difference_band > max_difference:
-                max_difference = max_difference_band
-            min_difference_band = np.min(is_close_differences)
-            if min_difference_band < min_difference:
-                min_difference = min_difference_band
-        else:
-            flag_bands_are_equal = True
+                print(prefix + f"     * image band {b} difference count is {num_image_differences} "
+                      f"with maximum difference of {max_difference_band} and minimum difference of "
+                      f"{min_difference_band}.")
+                differing_indices = np.where(~is_close_result)
+                i, j = differing_indices[0][0], differing_indices[1][0]
+                print(prefix + f'     * image band {b}, input 1 has value'
+                                  f' "{image_1[i, j]}" in position'
+                                  f' (i: {i}, j: {j})'
+                                  f' whereas input 2 has value "{image_2[i, j]}"'
+                                  ' in the same position.')
+            else:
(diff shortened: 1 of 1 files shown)


.ci/scripts/dswx_compare_opera_pge.py
```diff
@@ -0,0 +1,221 @@
+from osgeo import gdal
+import numpy as np
+import os
+import sys
+
+COMPARE_DSWX_HLS_PRODUCTS_ERROR_TOLERANCE_ATOL = 1e-6
+COMPARE_DSWX_HLS_PRODUCTS_ERROR_TOLERANCE_RTOL = 1e-5
+
+def _get_prefix_str(flag_same, flag_all_ok):
```
sjlewis-jpl, line 9: I think this was also in the original script.  I don't like changing input values as a surrogate for a return value, and would rather see a value for `flag_all_ok` returned instead.  But I cede to whatever the convention is for this repo.

sjlewis-jpl (changes requested):
Thank you for porting this over into our repos, I think having our own comparison script will be a good thing for the project in the long-run.

Most of my comments are in-line, including some changes to fully address the original ticket.

.ci/scripts/dswx_compare_opera_pge.py
```diff
@@ -0,0 +1,221 @@
...
+        num_image_differences = np.sum(~is_close_result)
+        if num_image_differences > 0:
+            flag_bands_are_equal = False
+            # Find the min and max differences using the numpy.isclose function
+            # 0 <= (atol + rtol * absolute(b)) - absolute(a-b)
+            a1 = image_1[~is_close_result]
+            b2 = image_2[~is_close_result]
+            atol = COMPARE_DSWX_HLS_PRODUCTS_ERROR_TOLERANCE_ATOL
+            rtol = COMPARE_DSWX_HLS_PRODUCTS_ERROR_TOLERANCE_RTOL
+            is_close_differences = abs((atol + rtol * abs(b2)) - abs(a1 - b2))
```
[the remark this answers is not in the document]
david-inglish: Is it possible that somebody would want to see the exact values that fell outside of the numpy.isclose() calculation in order to adjust the rtol and atol values and adjust the pass/fail tolerance of the comparison?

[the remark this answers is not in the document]
sjlewis-jpl: Not just possible, but highly likely that someone would do such an analysis.  Though the purpose should be on fixing the algorithm, rather than adjusting our criteria to get a passing test.  And that's easy enough to do offline, outside this script.

The main thing I was pointing out here is that, as written, `is_close_differences` aren't the actual differences, since the first term on the RHS adds a small value to the absolute differences.  You already know where the arrays are different, so just use that, rather than redoing the calculation behind `np.isclose()`.

Looking back at the code, and my suggestion up above, I might have gotten the meaning of `is_close_result` inverted.  Use the one that picks out where they're different (either `is_close_result` or `~is_close_result`).

.ci/scripts/dswx_compare_opera_pge.py
```diff
@@ -0,0 +1,221 @@
...
+                flag_same_metadata = False
+                metadata_error_message = (
+                    f'* the metadata key {k1} is present in'
+                    ' but it is not present in input 2')
+                break
+            # Exclude metadata fields that are not required to be the same
+            if k1 in ['PROCESSING_DATETIME', 'DEM_SOURCE', 'LANDCOVER_SOURCE',
+                      'WORLDCOVER_SOURCE']:
+                continue
+            if metadata_2[k1] != v1:
```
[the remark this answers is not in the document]
david-inglish: Added comment on 140

.ci/scripts/dswx_compare_opera_pge.py
```diff
@@ -0,0 +1,221 @@
...
+    else:
+        for k1, v1, in metadata_1.items():
+            if k1 not in metadata_2.keys():
+                flag_same_metadata = False
+                metadata_error_message = (
+                    f'* the metadata key {k1} is present in'
+                    ' but it is not present in input 2')
+                break
+            # Exclude metadata fields that are not required to be the same
+            if k1 in ['PROCESSING_DATETIME', 'DEM_SOURCE', 'LANDCOVER_SOURCE',
```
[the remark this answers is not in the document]
collinss-jpl: The main issue is that in PCM we sorta reserve the right to rename ancillary files to simplify setup of the runconfig template, so even a basename check would likely fail in most situations. If we were truly out of sync with the correct ancillaries to use, I would expect that to be caught during the array comparisons of the DEM/land/world cover layers.

.ci/scripts/dswx_compare_opera_pge.py
```diff
@@ -0,0 +1,223 @@
...
+        else:
+            # Currently these are string values otherwise the comparison below would need to change for floating point.
+            if metadata_2[k1] != v1:
+                msg = (f'* contents of metadata key {k1} from'
+                       f' input 1 has value "{v1}" whereas the same key in'
+                       f' input 2 metadata has value "{metadata_2[k1]}"')
+                # Don't fail for metadata fields that are not required to be the same
+                if k1 in metadata_exclude_list:
+                    # We will just print the difference in the output
+                    print(f'[INFO] {msg}')
```
sjlewis-jpl, line 196: These will print before the "Comparing metadata" message on line 140, which might be confusing.  I suggest just appending these into the `metadata_error_message` string, or making another return argument to contain the info messages, and print them all at the same time.  Probably good to group these together after the other metadata error messages.

.ci/scripts/dswx_compare_opera_pge.py
```diff
@@ -0,0 +1,223 @@
...
+    layer_gdal_dataset_2 = gdal.Open(file_2, gdal.GA_ReadOnly)
+    geotransform_2 = layer_gdal_dataset_2.GetGeoTransform()
+    metadata_2 = layer_gdal_dataset_2.GetMetadata()
+    nbands_2 = layer_gdal_dataset_2.RasterCount
+
+    # compare number of bands
+    flag_same_nbands =  nbands_1 == nbands_2
+    flag_all_ok, flag_same_nbands_str = _get_prefix_str(flag_same_nbands, flag_all_ok)
+    prefix = PREFIX
+    print(f'{flag_same_nbands_str}Comparing number of bands')
```
sjlewis-jpl, line 83: Add the prefix at the beginning of this line, to keep the first 7 characters of each line clear of anything other than a comparison status.

Or you can remove it.  If you put the error messages after the `"[FAIL] Band 1 - Water classification (WTR)"` line, you don't need this one.

.ci/scripts/dswx_compare_opera_pge.py
```diff
@@ -0,0 +1,223 @@
...
+                print(prefix +
+                      f'* e.g. image band {b} input 1 has value "{image_1[i, j]}" in position'
+                      f' (i: {i}, j: {j}) whereas input 2 has value "{image_2[i, j]}"'
+                      ' in the same position.')
+            else:
+                flag_bands_are_equal = True
+
+            flag_all_ok, flag_bands_are_equal_str = _get_prefix_str(flag_bands_are_equal,
+                                                       flag_all_ok)
+            print(f'{flag_bands_are_equal_str}     Band {b} -'
```
sjlewis-jpl, line 119: This has some extra spaces that mis-aligns the `"Band {b}"` part of the message.

Given my comment up above about removing the `"Comparing DSWx bands..."` line, replace those spaces with `"Comparing "`, so that it reads like one of these:
```
[FAIL] Comparing Band 1 - Water classification (WTR)
[OK]   Comparing Band 1 - Water classification (WTR)
```

At the very least, remove the extra extra spaces, so that `"Band {b}"` aligns with the rest of the output.

sjlewis-jpl (changes requested):
All of the issues with the actual comparisons are addressed well.

Every comparison has a "status" line starting with `'[OK].  '`, `'[FAIL] '`, or `'[INFO] '`, which is good.  When the comparison fails, we should be consistent about where the additional info goes - either always before or always right after (and before reporting failure details, it should say what is being compared).  I suggest having it go right after, as that saves a line identifying what is being reported on.  I made some specific suggestions in-line with the code to address this point.

These changes might seem like nit-picking, but since this will (hopefully) be used by multiple teams, and think having really clear and intuitive output is important.  

I attached two text files showing what the output looks like when a comparison succeeds vs fails, to see a concrete example of what I'm commenting on.  The failed output was on two different granules, so pretty much everything was wrong.

[comparison_failed_output.txt](https://github.com/nasa/opera-sds-pge/files/11032821/comparison_failed_output.txt)
[comparison_succeeded_output.txt](https://github.com/nasa/opera-sds-pge/files/11032823/comparison_succeeded_output.txt)

.ci/scripts/dswx_compare_opera_pge.py
david-inglish: I am committing changes to address the most-recent comments.
Here is sample output for a few scenarios with the new changes:
[comparison_all_ok.txt](https://github.com/nasa/opera-sds-pge/files/11065747/comparison_all_ok.txt)
[comparison_band_difference_fail.txt](https://github.com/nasa/opera-sds-pge/files/11065748/comparison_band_difference_fail.txt)
[comparison_fail_num_band_geotransform_metadata.txt](https://github.com/nasa/opera-sds-pge/files/11065749/comparison_fail_num_band_geotransform_metadata.txt)

sjlewis-jpl (approved):
The latest changes look good, and keep the spirit of my previous comments.  Nice work!

collinss-jpl (approved):
Thanks for all the hard work on this @david-inglish, and thanks again to @sjlewis-jpl for weighing in!

Issue #220, opened 2023-01-05
Rework metadata collection (via h5py)


collinss-jpl:
In the Beta version of the CSLC-S1 SAS, product metadata is now stored within a separate .h5 file, rather than .json.
We need to develop an analog of the the [get_rtc_s1_product_metadata](https://github.com/nasa/opera-sds-pge/blob/9aac542114419f7a52f2f2786cf1b41050c28590/src/opera/util/metadata_utils.py#L264) function tailored for the CSLC-S1 metadata format.

Similar to the RTC-S1 HDF5 metadata, there are also several top level groupings of metadata that we should capture into a dictionary. The definition of the available groupings can be found within section 5 of the CSLC-S1 Product Spec, which can be found in Artifactory-FN under the path `general/gov/nasa/jpl/opera/adt/cslc_s1/r3/beta/documents/OPERA_CSLC_S1_Product_Spec.docx`. The groupings to be captured should include:

- /science/SENTINEL1/identification
- /science/SENTNEL1/CSLC/grids
- /science/SENTINEL1/CSLC/corrections
- /science/SENTINEL1/CSLC/metadata/processing_information
- /science/SENTINEL1/CSLC/metadata/orbit

Sample .h5 products to test with can be found with the golden test dataset found in AF under the path `general/gov/nasa/jpl/opera/adt/cslc_s1/r3/beta/delivery_cslc_s1_beta_0.1.zip`

PR #226, opened 2023-01-17
220 Rework metadata collection via h5py

david-inglish:
## Description
- This adds a function to retrieve h5 metadata from the CSLC HDF5 metadata files.
- Testing showed that pruning datasets on retrieval doesn't save significant time and so pruning is not done.

## Affected Issues
- #220 

## Testing
- A unit test has been added to create dummy data which is retrieved and validated.


collinss-jpl (approved):
Looks great, thanks @david-inglish!

PR #231, opened 2023-01-30
DSWx-HLS Integration Test Asset Update

### LINKED ISSUE #230: DSWx-HLS v1.0.0-rc.6.0 Test Asset Update

collinss-jpl:
## Description
- This branch refactors the shared command-line argument interface used by all integration test scripts to allow the expected input and output archives to be specified as separate files. An additional argument to specify a custom location for the temporary directory has also been added to facilitate running the scripts in a non-Jenkins context.
- The integration script for the DSWx-HLS PGE has also been updated to accommodate the changes to shared interface
- The separate input/output test data assets have been uploaded to S3
- Made several small changes to reduce the output clutter within a Jenkins log

## Affected Issues
- Resolves #230 

## Testing
- Changes have been tested in Jenkins against the build/int test pipeline for the 219_dswx_int_test_asset_update branch

### DIFF
--- a/.ci/scripts/test_int_util.sh
+++ b/.ci/scripts/test_int_util.sh
@@ -5,6 +5,9 @@
 # Library of common shell routines for PGE integration testing.
 #
 
+# Base temp directory to use for file staging
+DEFAULT_TMP_ROOT="/data/tmp"
+
 test_int_parse_args()
 {
     # Parse command line arguments and set global variables to be used in calling script
@@ -12,22 +15,31 @@ test_int_parse_args()
     while [[ $# -gt 0 ]]; do
     case $1 in
       -h|--help)
-        echo "Usage: $(basename $0) [-h|--help] [-t|--tag <tag>] [--testdata <testdata .zip file>] [--runconfig <runconfig .yaml file>]"
+        echo "Usage: $(basename $0) [-h|--help] [-t|--tag <tag>] [-i|--input-data <zip file>] [-e|--expected-data <zip file>] [--runconfig <runconfig .yaml file>] [--temp-root <path>]"
         exit 0
         ;;
       -t|--tag)
         PGE_TAG=$2
-        echo "PGE_TAG is $PGE_TAG"
         shift
         shift
         ;;
-      --testdata)
-        TESTDATA=$2
+      -i|--input-data)
+        INPUT_DATA=$2
+        shift
+        shift
+        ;;
+      -e|--expected-data)
+        EXPECTED_DATA=$2
         shift
         shift
         ;;
       --runconfig)
         RUNCONFIG=$2
+        shift
+        shift
+        ;;
+      --temp-root)
+        TMP_ROOT=$2
         shift
         shift
         ;;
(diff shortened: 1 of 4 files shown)


.ci/scripts/util.sh
```diff
@@ -219,8 +219,11 @@ metrics_collection_end()
 
     # kill the background tasks (the pid number is stored in the file below)
     kill "$(cat "${stats_pid_file}")"
+    wait "$(cat "${stats_pid_file}")" 2> /dev/null || true
```
david-inglish, line 222: What mechanism will end the processes that were previously being killed here?

collinss-jpl: The `kill` on line 221 is still the mechanism that reaps the subproc, the `wait` was added so the stderr output about the killed process could be redirected to `/dev/null`. I lifted this approach from the answer to this thread: https://stackoverflow.com/questions/5719030/bash-silently-kill-background-function-process

david-inglish: I see, I just made a mistake in looking at this as a line replacement. Thanks!

david-inglish (approved):

.ci/scripts/test_int_util.sh
```diff
@@ -5,24 +5,31 @@
...
     # Parse command line arguments and set global variables to be used in calling script
     #
     while [[ $# -gt 0 ]]; do
     case $1 in
       -h|--help)
-        echo "Usage: $(basename $0) [-h|--help] [-t|--tag <tag>] [--testdata <testdata .zip file>] [--runconfig <runconfig .yaml file>]"
+        echo "Usage: $(basename $0) [-h|--help] [-t|--tag <tag>] [-i|--input-data <zip file>] [-e|--expected-data <zip file>] [--runconfig <runconfig .yaml file>] [--temp-root <path>]"
         exit 0
         ;;
       -t|--tag)
```
JimHofman, line 21: Although the arguments are pretty obvious, it might be useful to use comments to describe the arguments much at we do for our python code.

.ci/scripts/test_int_util.sh
```diff
@@ -68,53 +80,59 @@ test_int_setup_data_tmp_directory()
...
 
     # Pull in test data and runconfig from S3
-    echo "Downloading test data from s3://operasds-dev-pge/${PGE_NAME}/${TESTDATA} to $local_testdata_archive"
-    aws s3 cp s3://operasds-dev-pge/${PGE_NAME}/${TESTDATA} $local_testdata_archive --no-progress
-    echo "Downloading runconfig from s3://operasds-dev-pge/${PGE_NAME}/${RUNCONFIG} to $local_runconfig"
-    aws s3 cp s3://operasds-dev-pge/${PGE_NAME}/${RUNCONFIG} $local_runconfig
-
-    cd ${TMP_DIR}
+    echo "Downloading input data from s3://operasds-dev-pge/${PGE_NAME}/${INPUT_DATA} to $local_input_data_archive"
+    aws s3 cp s3://operasds-dev-pge/${PGE_NAME}/${INPUT_DATA} $local_input_data_archive --no-progress
```
JimHofman, line 97: When running locally, is there some setup required to get this to work?  If so maybe at a comment that points to the wiki (and of course a quick write up on the wiki)

JimHofman (approved):

PR #236, opened 2023-02-06
232 CSLC-S1 v2.0.0-rc.5.0 Test Asset Update

david-inglish:
## Description
- Test assets have been updated to be compatible with https://github.com/nasa/opera-sds-pge/wiki/CSLC-S1-Beta-Acceptance-Testing-Instructions

## Affected Issues
- #232 

## Testing
- Build/Integration pipeline passes for cslc_s1.

### DIFF
--- a/.ci/scripts/test_int_cslc_s1.sh
+++ b/.ci/scripts/test_int_cslc_s1.sh
@@ -1,6 +1,7 @@
 #!/bin/bash
 # Script to execute integration tests on OPERA CSLC_S1 PGE Docker image
 #
+set -e
 umask 002
 
 SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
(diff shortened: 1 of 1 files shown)


.ci/scripts/test_int_cslc_s1.sh
```diff
@@ -52,30 +52,32 @@ aws s3 cp s3://operasds-dev-pge/${PGE_NAME}/validate_cslc.py "$local_validate_cs
 # 0 - pass
 # 1 - failure to execute some part of this script
 # 2 - product validation failure
-overall_status=0
+overall_status=1
```
JimHofman, line 55: I'm interested to know why you initialized to failure.

david-inglish: Looking at this again, it doesn't seem necessary but it made it clear to me that the script would not return a "pass" status unless the PGE execution returned a zero status (where overall_status is now set to 0).

I removed the "set -e" line to allow the script to get through python exceptions I was seeing during the comparison step, and that opened the possibility of getting further in the script after earlier things failed.

I now think a slightly better approach would be to restore the "set -e" and run the comparisons in a way that captures the return values from the comparison step without exiting the script via "set -e" as described here: https://stackoverflow.com/questions/18621990/bash-get-exit-status-of-command-when-set-e-is-active

JimHofman (approved):

.ci/scripts/test_int_cslc_s1.sh
collinss-jpl: I like this approach you linked, and would like to keep the `set -e` in place

collinss-jpl (approved):

PR #241, opened 2023-02-17
238 flake pytest

JimHofman:
## Description
- Cleaned up python files so we have a clean flake8 run
- added **_per-file-ignores = metadata_utils.py: F841_** to .flake8 file
## Affected Issues
- N/A

## Testing
- ran coverage test with all unit tests passing


.flake8
```diff
@@ -2,6 +2,7 @@
 format = pylint
 output-file = flake8.log
 max-line-length = 120
+per-file-ignores = metadata_utils.py: F841
```
collinss-jpl, line 5: 👍 nice find, didn't know you could configure this here

collinss-jpl (approved):
Cleanup looks great, thanks @JimHofman!

PR #243, opened 2023-02-28
239 code coverage

JimHofman:
## Description
- Added to test file to get all coverage over  90%

## Testing
- Ran the test suite successfully

### DIFF
--- a/src/opera/pge/rtc_s1/rtc_s1_pge.py
+++ b/src/opera/pge/rtc_s1/rtc_s1_pge.py
@@ -470,6 +470,21 @@ class RtcS1PostProcessorMixin(PostProcessorMixin):
         output_product_metadata['frequencyA']['frequencyALength'] = len(output_product_metadata['frequencyA']
                                                                         ['yCoordinates'])
 
+        # TODO: the following fields seems to be missing in the interface delivery products,
+        #       but are documented, remove these kludges once they are actually available
+        if 'azimuthBandwidth' not in output_product_metadata['frequencyA']:
+            output_product_metadata['frequencyA']['azimuthBandwidth'] = 12345678.9
+
+        if 'noiseCorrectionFlag' not in output_product_metadata['frequencyA']:
+            output_product_metadata['frequencyA']['noiseCorrectionFlag'] = False
+
+        if 'plannedDatatakeId' not in output_product_metadata['identification']:
+            output_product_metadata['identification']['plannedDatatakeId'] = ['datatake1', 'datatake2']
+
+        if 'plannedObservationId' not in output_product_metadata['identification']:
+            output_product_metadata['identification']['plannedObservationId'] = ['obs1', 'obs2']
+        # TODO: end kludges
+
         return output_product_metadata
 
     def _create_custom_metadata(self):
--- a/src/opera/pge/rtc_s1/rtc_s1_pge.py
+++ b/src/opera/pge/rtc_s1/rtc_s1_pge.py
@@ -472,17 +472,17 @@ class RtcS1PostProcessorMixin(PostProcessorMixin):
 
         # TODO: the following fields seems to be missing in the interface delivery products,
         #       but are documented, remove these kludges once they are actually available
-        if 'azimuthBandwidth' not in output_product_metadata['frequencyA']:
-            output_product_metadata['frequencyA']['azimuthBandwidth'] = 12345678.9
-
-        if 'noiseCorrectionFlag' not in output_product_metadata['frequencyA']:
-            output_product_metadata['frequencyA']['noiseCorrectionFlag'] = False
-
-        if 'plannedDatatakeId' not in output_product_metadata['identification']:
-            output_product_metadata['identification']['plannedDatatakeId'] = ['datatake1', 'datatake2']
-
-        if 'plannedObservationId' not in output_product_metadata['identification']:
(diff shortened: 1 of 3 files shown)


src/opera/pge/rtc_s1/rtc_s1_pge.py
```diff
@@ -465,23 +465,10 @@ def _collect_rtc_product_metadata(self, metadata_product):
         output_product_metadata = get_rtc_s1_product_metadata(metadata_product)
 
         # Fill in some additional fields expected within the ISO
-        output_product_metadata['frequencyA']['frequencyAWidth'] = len(output_product_metadata['frequencyA']['xCoordinates'])
-        output_product_metadata['frequencyA']['frequencyALength'] = len(output_product_metadata['frequencyA']['yCoordinates'])
-
-        # TODO: the following fields seems to be missing in the interface delivery products,
```
collinss-jpl, line 471: I don't think this section between the `TODOs` should have been removed.

src/opera/test/pge/cslc_s1/test_cslc_s1_pge.py
```diff
@@ -77,6 +77,16 @@ def setUp(self) -> None:
             f"touch {join(input_dir, 'db.sqlite3')}"
         )
 
+        os.system(
+            f"touch {join(input_dir, 'test_qa.py')}"
+        )
+        os.chmod(f"{join(input_dir, 'test_qa.py')}", stat.S_IRWXU)   # Set to read, write, execute by owner
```
collinss-jpl, line 83: So its a little clearer what the purpose of these dummy python scripts are, can we change the name to `test_qa_rwx.py` for the version that has read/write/execute privileges, and set the other to `test_qa_ro.py` for the read-only version?

src/opera/test/pge/cslc_s1/test_cslc_s1_pge.py
```diff
@@ -77,6 +77,16 @@ def setUp(self) -> None:
             f"touch {join(input_dir, 'db.sqlite3')}"
         )
 
+        os.system(
```
collinss-jpl, line 80: It would be nice to have a comment before these files that explains that they're intended for use with the `test_qa_enabled` function

src/opera/test/pge/cslc_s1/test_cslc_s1_pge.py
```diff
@@ -212,6 +223,35 @@ def test_iso_metadata_creation(self):
...
+            with self.assertRaises(RuntimeError):
+                pge.run()
+
+            expected_log_file = pge.logger.get_file_name()
+            self.assertTrue(os.path.exists(expected_log_file))
+
+            with open(expected_log_file, 'r', encoding='utf-8') as infile:
+                log_contents = infile.read()
+
+            self.assertIn("Could not load ISO template", log_contents)
```
collinss-jpl, line 250: 👍

src/opera/test/pge/cslc_s1/test_cslc_s1_pge.py
```diff
@@ -346,6 +386,72 @@ def test_cslc_s1_output_validation(self):
...
+    def test_geotiff_json_filenames(self):
+        """Test that tiff and json filenames are properly returned."""
+        runconfig_path = join(self.data_dir, 'test_cslc_s1_config.yaml')
+        pge = CslcS1Executor(pge_name="CslcPgeTest", runconfig_path=runconfig_path)
+        pge.run()
+        inner_fname = 'cslc_pge_test/output_dir/t064_135518_iw1/20220501/t064_135518_iw1_20220501_VV.tiff'
+        pge._geotiff_filename(inner_fname)
+        inner_fname = 'cslc_pge_test/output_dir/t064_135518_iw1/20220501/t064_135518_iw1_20220501_VV.json'
+        pge._json_metadata_filename(inner_fname)
+        pge._qa_log_filename()
```
collinss-jpl, line 403: I understand that this test is probably providing some additional coverage, but I don't actually see it "testing" anything, so its a little hard to follow. Is there an `assertEquals` that could be added? If the point of the test is just to make sure that an exception is not raised, then we could add a try/except block, where the except block call `self.fail()`, so its clear we do not want it to be hit.

src/opera/test/pge/rtc_s1/test_rtc_s1_pge.py
```diff
@@ -67,6 +68,14 @@ def setUp(self) -> None:
 
         os.system(f"touch {join(input_dir, 'db.sqlite3')}")
 
+        os.system(f"touch {join(input_dir, 'test_qa.py')}"
```
collinss-jpl, line 71: Similar comment as above regarding rename of test files here.

collinss-jpl (changes requested):
The new tests look great, but I had some suggestions for a few of them to make their purpose a little more clear. Once they're addressed this will be good to merge.

src/opera/pge/rtc_s1/rtc_s1_pge.py
JimHofman: I put them back in, but I removed them because all the fields were found so I assumed that those specific fields were now available.

collinss-jpl (approved):
Changes all look good, went ahead and rebased and resolved merged conflicts for this branch. As soon as it has passed the CI checks I will merge. Thanks @JimHofman!

Issue #257, opened 2023-03-27
DSWx-S1 Initial Interface Integration


collinss-jpl:
The existing stub implementation of the DSWx-S1 PGE needs to be updated to integrate with the now-delivered interface SAS. The goal of this ticket is to add all the scripts/configs needed to build a combined PGE/SAS container for DSWx-S1. This includes the following tasks:

- [x] Add the [DSWx-S1 SAS schemas](https://github.com/opera-adt/DSWX-SAR/tree/main/src/dswx_sar/schemas) (both of them) into the `src/opera/pge/dswx_s1/schema` directory
- [x] Create the `Dockerfile_dswx_s1` file within the `.ci/docker directory`. This file should just be a copy of the `Dockerfile_dswx_hls` to begin with.
- [x] Create the `build_dswx_s1.sh` and `test_dswx_s1.sh` scripts under `.ci/scripts`. These can be based on the corresponding version for DSWx-HLS, but will need some tailoring to make them applicable to DSWx-S1 (for example updating the artifactory path to the SAS docker base image)
- [x] Update the `build_all_images.sh` and `test_all_images.sh` scripts (also under `.ci/scripts`) to include the new scripts added in the previous task
- [x] Create an example runconfig for the DSWx-S1 PGE named `dswx_s1_sample_runconfig-v3.0.0-er.1.0.yaml` and populate it with both the PGE and SAS sections of the Runconfig. The sample SAS runconfig provided with the Interface delivery can be used as the example SAS portion.

PR #261, opened 2023-04-03
#257 - initial dswx_s1

JimHofman:
## Description
- Initial stub implementations of DSWX-S1 PGE for the interface SAS
## Affected Issues
- N/A

## Testing
Successfully ran test_dsxw_s1.sh verifying interface to the dswx_s1 SAS

### DIFF
--- a/examples/dswx_s1_sample_runconfig-v3.0.0-er.1.0.yaml
+++ b/examples/dswx_s1_sample_runconfig-v3.0.0-er.1.0.yaml
@@ -1,0 +1,200 @@
+# Sample RunConfig for use with the DSWx-S1 PGE
+# This RunConfig should require minimal changes in order to be used with the
+# OPERA PCM.
+
+RunConfig:
+    # Name for the RunConfig, may be any string
+    Name: OPERA-DSWX-S1-PGE-SAMPLE-CONFIG
+
+    Groups:
+        # PGE-specific RunConfig section
+        # This section is only used by the PGE, however, paths to inputs/outputs
+        # should align with the similar sections of the SAS RunConfig
+        PGE:
+            PGENameGroup:
+                # Name of the PGE for use with this RunConfig, should always be
+                # DSWX_S1_PGE when using with the DSWx-S1 PGE
+                PGEName: DSWX_S1_PGE
+
+            InputFilesGroup:
+                # List of input files/directories
+                # Must be either a top-level directory containing one or more HLS
+                # .tif files, or a list of all HLS .tif files for use with the job
+                # Paths must correspond to the file system within the Docker container
+                InputFilePaths:
+                    - /home/conda/input_dir
+
+            DynamicAncillaryFilesGroup:
+                # Map of ancillary file types to paths to the file
+                # Paths must correspond to the file system within the Docker container
+                AncillaryFileMap:
+                    # Path to the Digital Elevation Model
+                    # Must be either a single .tif, or a .vrt that points to one
+                    # or more .tif files
+                    dem_file: /home/conda/input_dir/dem.tif
+
+                    # Path to the Discrete Classification Map
+                    # Must point to a single .tif file
+                    hand_file: /home/conda/input_dir/hand.tif
+
+                    reference_water_file: /home/conda/input_dir/reference_water.tif
+
+                    worldcover: /home/conda/input_dir/worldcover.tif
+
+
+            ProductPathGroup:
+                # Path to where output products should be stored
+                # Must correspond to the file system within the Docker container,
+                # and must have write permissions for the User/Group used with
+                # the "Docker run" command
+                OutputProductPath: /home/conda/output_dir
+
+                # Path to a scratch directory for the PGE and SAS to store
+                # intermediate files that will not be needed after PGE execution
+                # completes
+                # Must correspond to the file system within the Docker container,
+                # and must have write permissions for the User/Group used with
+                # the "Docker run" command
+                ScratchPath: /home/conda/output_dir/scratch_dir
+
+            PrimaryExecutable:
+                # Identifier for the PGE executable, should always be DSWX_HLS for
+                # this PGE
+                ProductIdentifier: DSWX_S1
+
+                # Product version specific to output products
+                ProductVersion: 1.0
+
+                # Path to the executable to run, path must be reachable from
+                # within the Docker container (i.e. findable with a "which" command)
+                ProgramPath: python3
+
+                # List of command-line options to use with ProgramPath
+                ProgramOptions:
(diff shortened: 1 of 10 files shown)


examples/dswx_s1_sample_runconfig-v3.0.0-er.1.0.yaml
```diff
@@ -0,0 +1,200 @@
...
+        # should align with the similar sections of the SAS RunConfig
+        PGE:
+            PGENameGroup:
+                # Name of the PGE for use with this RunConfig, should always be
+                # DSWX_S1_PGE when using with the DSWx-S1 PGE
+                PGEName: DSWX_S1_PGE
+
+            InputFilesGroup:
+                # List of input files/directories
+                # Must be either a top-level directory containing one or more HLS
```
collinss-jpl, line 21: This part of the description looks to be leftover from the DSWx-HLS example, since it references HLS files. It should instead say something like "Must be a list of directories, each of which contains a single RTC burst product .tif and .h5 metadata file"

examples/dswx_s1_sample_runconfig-v3.0.0-er.1.0.yaml
```diff
@@ -0,0 +1,200 @@
...
+                    # or more .tif files
+                    dem_file: /home/conda/input_dir/dem.tif
+
+                    # Path to the Discrete Classification Map
+                    # Must point to a single .tif file
+                    hand_file: /home/conda/input_dir/hand.tif
+
+                    reference_water_file: /home/conda/input_dir/reference_water.tif
+
+                    worldcover: /home/conda/input_dir/worldcover.tif
```
collinss-jpl, line 42: The key should match what the SAS uses, for consistency. So this should be `worldcover_file`

examples/dswx_s1_sample_runconfig-v3.0.0-er.1.0.yaml
```diff
@@ -0,0 +1,200 @@
...
+                # DSWX_S1_PGE when using with the DSWx-S1 PGE
+                PGEName: DSWX_S1_PGE
+
+            InputFilesGroup:
+                # List of input files/directories
+                # Must be either a top-level directory containing one or more HLS
+                # .tif files, or a list of all HLS .tif files for use with the job
+                # Paths must correspond to the file system within the Docker container
+                InputFilePaths:
+                    - /home/conda/input_dir
```
collinss-jpl, line 25: Lets set this to same set of values used by the SAS portion of the runconfig

examples/dswx_s1_sample_runconfig-v3.0.0-er.1.0.yaml
```diff
@@ -0,0 +1,200 @@
...
+                # Product version specific to output products
+                ProductVersion: 1.0
+
+                # Path to the executable to run, path must be reachable from
+                # within the Docker container (i.e. findable with a "which" command)
+                ProgramPath: python3
+
+                # List of command-line options to use with ProgramPath
+                ProgramOptions:
+                    - /home/conda/dswx-s1-1.0.1/bin/dswx_s1.py
```
collinss-jpl, line 74: This should be `/home/conda/dswx-s1-0.1.0/src/dswx_sar/dswx_s1.py`

examples/dswx_s1_sample_runconfig-v3.0.0-er.1.0.yaml
```diff
@@ -0,0 +1,200 @@
...
+                    - --full-log-format
```
collinss-jpl, line 75: The SAS may not support this argument yet, so it should be removed for now

examples/dswx_s1_sample_runconfig-v3.0.0-er.1.0.yaml
```diff
@@ -0,0 +1,200 @@
...
+                # Consult the Docker image build scripts for more info
+                SchemaPath: /home/conda/opera/pge/dswx_s1/schema/dswx_s1_sas_schema.yaml
+
+                # Path to the Jinja2 template used to generate the ISO xml
+                # metadata file
+                # Path should correspond to the file system within the Docker
+                # container, and typically should reference a template file bundled
+                # with the opera_pge installation directory within the container
+                # Consult the Docker image build scripts for more info
+                IsoTemplatePath: /home/conda/opera/pge/dswx_s1/templates/OPERA_ISO_metadata_L3_DSWx_S1_template.xml.jinja2
```
collinss-jpl, line 95: Since this template doesn't actually exist yet, we should leave this field blank. Otherwise running with this example config could cause a file not found error. If its left empty, I believe the ISO step will be skipped.

examples/dswx_s1_sample_runconfig-v3.0.0-er.1.0.yaml
```diff
@@ -0,0 +1,200 @@
...
+              pge_name_group:
+                pge_name: DSWX_S1_PGE
+
+              input_file_group:
+                # REQUIRED - Single S1 RTC geotiff or
+                #            multiple burst RTC directory containing
+                #            hdf5 metadata and rtc geotiff (directory or files)
+                #            for open water
+                input_file_path: [
+                  input_data/rtc_data/t047_100908_iw3,
```
collinss-jpl, line 136: For the paths in the SAS section, let's be consistent about using absolute paths, for example `/home/conda/input_data/rtc_data/...`

collinss-jpl (changes requested):
Everything looks good, just the example runconfig needs a few touchups. Great work otherwise @JimHofman

examples/dswx_s1_sample_runconfig-v3.0.0-er.1.0.yaml
JimHofman: Nice catch Scott.

collinss-jpl: Looks like this got updated to `world_cover`, and not `worldcover_file`...

JimHofman: I changed it, but it looks like it's worldcover_file in dsw_s1.yaml from the download.

JimHofman: I also changed it in the schema file

collinss-jpl: Ah I see, I had not realized that the SAS schema was also calling this filed "world_file", thanks for addressing

collinss-jpl (approved):

PR #269, opened 2023-04-19
266 dswx s1 schema val

JimHofman:
## Description
- Ticket 266:  DSWx-S1 algorithm schema validation

## Testing
- made new unit tests for the new code

### DIFF
--- a/src/opera/pge/base/runconfig.py
+++ b/src/opera/pge/base/runconfig.py
@@ -85,6 +85,34 @@ class RunConfig:
                 f'Unable to parse {yaml_filename}, expected top-level RunConfig entry'
             ) from key_error
 
+    @staticmethod
+    def _parse_algorithm_parameters_run_config_file(yaml_filename):
+        """
+        Loads an algorithm parameter run configuration YAML file.
+        Returns the loaded data as a Python object.
+
+        Parameters
+        ----------
+        yaml_filename : str
+            Path to the runconfig YAML file to parse.
+
+
+        Raises
+        ------
+        RuntimeError
+            If the parsed config does not define a top-level "runconfig" entry
+
+        """
+        with open(yaml_filename, 'r', encoding='utf-8') as stream:
+            dictionary = yaml.safe_load(stream)
+
+        try:
+            return dictionary['runconfig']
+        except KeyError as key_error:
+            raise RuntimeError(
+                f'Unable to parse {yaml_filename}, expected top-level runconfig entry'
+            ) from key_error
+
     def validate(self, pge_schema_file=BASE_PGE_SCHEMA, strict_mode=True):
         """
         Validates the RunConfig using a combination of the base PGE schema,
@@ -250,6 +278,27 @@ class RunConfig:
             sas_schema_path
             if isabs(sas_schema_path)
             else resource_filename('opera', sas_schema_path)
+        )
+
+    @property
+    def algorithm_parameters_schema_path(self) -> str:
+        """Returns the path to the algorithm parameter Schema file for DSWX-S1"""
+        algorithm_parameters_schema_path = self._pge_config['PrimaryExecutable']['AlgorithmParametersSchemaPath']
+        return (
+            algorithm_parameters_schema_path
+            if isabs(algorithm_parameters_schema_path)
+            else resource_filename('opera', algorithm_parameters_schema_path)
+        )
+
+    @property
+    def algorithm_parameters_config_path(self) -> str:
+        """Returns the path to the optional algorithm_parameter run configuration file for DSWX-S1"""
+        algorithm_parameters_config_path = \
+            self._sas_config['runconfig']['groups']['dynamic_ancillary_file_group']['algorithm_parameters']
+        return (
+            algorithm_parameters_config_path
+            if isabs(algorithm_parameters_config_path)
+            else resource_filename('opera', algorithm_parameters_config_path)
         )
 
     @property
--- a/src/opera/pge/base/runconfig.py
+++ b/src/opera/pge/base/runconfig.py
@@ -282,24 +282,30 @@ class RunConfig:
 
     @property
     def algorithm_parameters_schema_path(self) -> str:
-        """Returns the path to the algorithm parameter Schema file for DSWX-S1"""
+        """Returns the path to the optional algorithm parameter Schema file for DSWX-S1"""
         algorithm_parameters_schema_path = self._pge_config['PrimaryExecutable']['AlgorithmParametersSchemaPath']
-        return (
-            algorithm_parameters_schema_path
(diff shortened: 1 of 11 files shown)


src/opera/pge/base/runconfig.py
```diff
@@ -85,6 +85,34 @@ def _parse_run_config_file(yaml_filename):
...
+        """
+        with open(yaml_filename, 'r', encoding='utf-8') as stream:
+            dictionary = yaml.safe_load(stream)
+
+        try:
+            return dictionary['runconfig']
+        except KeyError as key_error:
+            raise RuntimeError(
+                f'Unable to parse {yaml_filename}, expected top-level runconfig entry'
+            ) from key_error
```
collinss-jpl, line 114: 👍

src/opera/pge/base/runconfig.py
```diff
@@ -252,6 +280,27 @@ def sas_schema_path(self) -> str:
...
+            algorithm_parameters_schema_path
+            if isabs(algorithm_parameters_schema_path)
+            else resource_filename('opera', algorithm_parameters_schema_path)
+        )
+
+    @property
+    def algorithm_parameters_config_path(self) -> str:
+        """Returns the path to the optional algorithm_parameter run configuration file for DSWX-S1"""
+        algorithm_parameters_config_path = \
+            self._sas_config['runconfig']['groups']['dynamic_ancillary_file_group']['algorithm_parameters']
```
collinss-jpl, line 297: Since not all SAS utilize a separate algorithm parameters run config, this method needs to be a bit sensitive to that. 

We should wrap this function body with a try/except block that looks for `KeyError` (i.e. when there is no `algorithm_parameters` field in the parsed runconfig), and if we hit the except block we can assume the field is not present and just return `None` instead.

src/opera/pge/dswx_s1/dswx_s1_pge.py
```diff
@@ -27,6 +32,46 @@ class DSWxS1PreProcessorMixin(PreProcessorMixin):
...
+        and another for algorithm parameters.  This allows for independent modification
+        of algorithm parameters within it's own runconfig file. This method performs validation
+        of the 'algorithm parameters' runconfig file against its associated schema file. The SAS
+        section of the main runconfig defines the location within the container of the 'algorithm
+        parameters' runconfig file, under ['dynamic_ancillary_file_group']['algorithm_parameters'].
+        The schema file for the 'algorithm parameters' runconfig file is referenced under
+        ['PrimaryExecutable']['ParametersSchemaPath'] in the PGE section of the runconfig file.
+
+        """
+        self.runconfig = RunConfig(self.runconfig_path)
```
collinss-jpl, line 47: Since `DSWxS1PreProcessorMixin.run_preprocessor` runs this function after the pre-processor activity of the base PGE, I'm not sure this line is necessary. This method should be able to assume the "main" RunConfig has already been parsed, validated and assigned to `self.runconfig`

src/opera/pge/base/runconfig.py
```diff
@@ -252,6 +280,27 @@ def sas_schema_path(self) -> str:
             else resource_filename('opera', sas_schema_path)
         )
 
+    @property
+    def algorithm_parameters_schema_path(self) -> str:
+        """Returns the path to the algorithm parameter Schema file for DSWX-S1"""
+        algorithm_parameters_schema_path = self._pge_config['PrimaryExecutable']['AlgorithmParametersSchemaPath']
+        return (
```
collinss-jpl, line 287: Since `AlgorithmParametersSchemaPath` has been added as an optional field, this method needs to account for the case where no value was provided at all. As is, a value of `None` would trigger the `resource_filename` call, which is not what we want. I think this method needs to check `algorithm_parameters_schema_path` is not `None`, and only then should it move on to the `return` logic.

If `algorithm_parameters_schema_path` is `None`, then that should be returned and its up to the caller to check if it got a result back.

src/opera/pge/dswx_s1/dswx_s1_pge.py
```diff
@@ -27,6 +32,46 @@ class DSWxS1PreProcessorMixin(PreProcessorMixin):
...
+                f'Can not validate algorithm_parameters schema file.  '
+                f'File: ({algorithm_parameters_schema_file_path}) not found.'
+            )
+
+        # Get the 'algorithm parameters' runconfig file
+        self.algorithm_parameters_runconfig = self.runconfig.algorithm_parameters_config_path
+        if isfile(self.algorithm_parameters_runconfig):
+            # Load the 'algorithm parameters' runconfig file
+            algorithm_parameters_config_data = yamale.make_data(self.algorithm_parameters_runconfig)
+            pass
```
collinss-jpl, line 65: This `pass` should be removed

src/opera/pge/dswx_s1/schema/dswx_s1_sas_schema.yaml
```diff
@@ -65,4 +65,4 @@ runconfig:
 
             product_version: num(required=False)
 
-        log_file: 
\ No newline at end of file
+        log_file: str()
```
collinss-jpl, line 68: We need to keep this file in sync with the version ADT maintains, so this should be set to `str(required=False)`

src/opera/test/data/test_dswx_s1_config.yaml
```diff
@@ -25,8 +28,9 @@ RunConfig:
                 ProgramOptions:
                     - 'hello world > dswx_s1_pge_test/output_dir/dswx_s1_test.txt;'
                     - '/bin/echo DSWx-S1 invoked with RunConfig'
-                ErrorCodeBase: 200000
+                ErrorCodeBase: 100000
```
collinss-jpl, line 31: Not critical, but this is the error code base used by DSWx-HLS. The idea was to have each PGE use a unique base value, and I've incremented by `100000` for each new PGE, so the "official" offset for DSWx-S1 should be `400000`

collinss-jpl (changes requested):
Overall, the updates look great as regards the core validation code. However, I had a few findings to make use of the algo parameter fields in the RunConfig a bit more robust, since not all PGE/SAS will have them set.

src/opera/pge/dswx_s1/dswx_s1_pge.py
JimHofman: removed

src/opera/pge/dswx_s1/dswx_s1_pge.py
JimHofman: oops, hangover from when I was first setting things up

collinss-jpl: Looks like this is still present after your latest changes

collinss-jpl (approved):

Issue #281, opened 2023-05-08
DISP-S1 algorithm parameter schema validation


collinss-jpl:
Similar to DSWx-S1, the DISP-S1 SAS has also added an externalized algorithm parameters yaml config which is treated as an "ancillary" input to the workflow. The DISP-S1 PGE needs to add a step to its pre-processor class to perform Yamale validation on this separate config, after the primary RunConfig has been validated.

Tasks include:

- [x] Updated example/test RunConfigs for DISP-S1 to include the new `AlgorithmParametersSchemaPath` field within the `PrimaryExecutable` section. This field should be set to the path to the algorithim parameters yamale schema for DISP-S1
- [x] The existing `_validate_algorithm_parameters_config()` that was written for the DSWx-S1 pre-processor should also be sufficient for DISP-S1 (or any other PGE with an algo params config going forward), so the task should be to refactor this function into a utility within `input_validation.py`, then update both R3 PGE's to call the function in their respective pre-processors.
- [x] Add unit tests to the DISP-S1 suite for valid/invalid validation

collinss-jpl:
Merged with #286

PR #286, opened 2023-05-16
281 DISP-S1 algorithm parameter schema validation

david-inglish:
## Description
- Updated example/test RunConfigs for DISP-S1 to include the new AlgorithmParametersSchemaPath field within the PrimaryExecutable section. This field should be set to the path to the algorithim parameters yamale schema for DISP-S1
- The existing _validate_algorithm_parameters_config() that was written for the DSWx-S1 pre-processor should also be sufficient for DISP-S1 (or any other PGE with an algo params config going forward), so the task should be to refactor this function into a utility within input_validation.py, then update both R3 PGE's to call the function in their respective pre-processors.
- Add unit tests to the DISP-S1 suite for valid/invalid validation

## Affected Issues
- #281 
- 
## Testing
- Added unit tests

### DIFF
--- a/src/opera/pge/disp_s1/schema/algor_param_disp_s1_schema.yaml
+++ b/src/opera/pge/disp_s1/schema/algor_param_disp_s1_schema.yaml
@@ -19,11 +19,11 @@ runconfig:
             # For single-reference network: Index of the reference image in the network.
             reference_idx: int()
             # Max `n` to form the nearest-`n` interferograms by index.
-            max_bandwidth: int()
+            max_bandwidth: any(int(), null())
             # Maximum temporal baseline of interferograms.
-            max_temporal_baseline: int()
+            max_temporal_baseline: any(int(), null())
             # For manual-index network: List of (ref_idx, sec_idx) defining the interferograms to form.
-            indexes: list()
+            indexes: list(list())
             # Type of interferogram network to create from phase-linking results.
             # Options: ['single-reference', 'manual-index', 'max-bandwidth', 'max-temporal-baseline'].
             network_type: enum('single-reference', 'manual-index', 'max-bandwidth', 'max-temporal-baseline')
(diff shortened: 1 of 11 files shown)


src/opera/pge/disp_s1/schema/algor_param_disp_s1_schema.yaml
```diff
@@ -39,15 +39,11 @@ runconfig:
...
             # Alternative to specifying output resolution: Specify the (x, y) strides (decimation
             #   factor) to perform while processing input. For example, strides of [4, 2] would turn an
             #   input resolution of [5, 10] into an output resolution of [20, 20].
-            strides:
-                x: int(required=False)
-                y: int(required=False)
+            strides: any(include('x_y_object'), null())
             # Options for `create_dataset` with h5py.
             hdf5_creation_options:
                 chunks: list()
```
JimHofman, line 49: Just double checking is this without parameters because it is a single list and 'indexes' is a list of lists?

src/opera/test/pge/disp_s1/test_disp_s1_pge.py
```diff
@@ -118,6 +150,104 @@ def test_disp_s1_pge_execution(self):
 
         self.assertIn(f"DISP-S1 invoked with RunConfig {expected_sas_config_file}", log_contents)
 
+    def test_disp_s1_pge_validate_algorithm_parameters_config(self):
```
JimHofman, line 153: My initial unittests for DSWx_S1 passed with similar code.  As things got more complex I ended up having to copy the algorithm_parameters runconfig into the input_dir within my temporary test environment, then modify the path in the test config file in the test/data directory.  That solved some weird behavior that I was having.  Just a heads up for when you start validating the inputs.

src/opera/util/metadata_utils.py
```diff
@@ -403,8 +403,9 @@ def get_cslc_s1_product_metadata(file_name):
         'identification': get_hdf5_group_as_dict(file_name, f"{S1_SLC_HDF5_PREFIX}/identification"),
         'grids': get_hdf5_group_as_dict(file_name, f"{S1_SLC_HDF5_PREFIX}/CSLC/grids"),
         'corrections': get_hdf5_group_as_dict(file_name, f"{S1_SLC_HDF5_PREFIX}/CSLC/corrections"),
-        'calibration_information': get_hdf5_group_as_dict(file_name,
-                                                          f"{S1_SLC_HDF5_PREFIX}/CSLC/metadata/calibration_information"),
+        'calibration_information':
```
JimHofman, line 406: Nice, flake8 thanks you.

JimHofman (approved):

src/opera/pge/disp_s1/schema/algor_param_disp_s1_schema.yaml
david-inglish: I believe that is correct. Indexes is a list of lists.  There are some comments that hint at what the correct schema construct should be but some of this is a best-guess at this point.

collinss-jpl (approved):
Great work as always @david-inglish. I made a few changes to resolve the merge conflicts and consolidate the two versions of the property method for getting the algorithm parameters file path from the SAS section of the runconfig. Once this branch has passed the CI system, I will merge it.

Issue #284, opened 2023-05-15
DISP-S1 Input/Ancillary Product Validation


collinss-jpl:
Ticket to track implementation of the validation functions for the inputs to the DISP-S1 PGE.

The input products for DISP-S1 can be classified into two groups: the main input products (the CSLC burst products) and the ancillary input products (DEM, geometry, amplitude mean/distortion, water mask, tec files, and weather model). This ticket proposes the following approaches for each type:

**CSLC Inputs**

- [x] Using the list of inputs in the `cslc_file_list` section of the SAS runconfig, verify that each file exists, is non-zero in size, and ends with the expected extension (`.h5`). 
- [ ] _Stretch goal_: while iterating over the list of inputs, compile a set of the unique burst ID's across all products. This will be used in the ancillary input validation, so the result should be assigned to the PGE object instance.

Note that (for now at least), the compressed and non-compressed CSLC files can be treated identically.

**Ancillary inputs**

- [x] DEM file: confirm existence, non-zero size, and ends with `.tif` or `.tiff` extension
- [x] Geometry: verify all files in list of `geometry_files` exist, are non-zero in size, and end with expected `.h5` extension
- [x] Amplitude mean/distortion: verify all files in the `amplitude_dispersion_files` and `amplitude_mean_files` exist, are non-zero  and end with expected `.tif` or `.tiff` extension
- [x] Water mask: confirm existence, non-zero size, and ends with `.tif` or `.tiff` extension
- [x] TEC files: verify all files in list of `tec_files` exist, and are non-zero in size. **No extension checking for this file type**.
- [x] Weather model files: verify all files in list of `weather_model_files` exist, are non-zero in size, and end with expected `.nc` or `.h5` extensions
- [ ]  _Stretch goal_: using the list of unique input burst ID's generated previously, add steps to validate that the amplitude mean/dispersion and geometry files line up with respect to number of inputs (one per unique burst), as well as the set of burst unique burst ID's.

Unit tests should be provided in test_disp_s1_pge.py to test the new validation functionality.

PR #292, opened 2023-06-01
284 DISP-S1 input/ancillary product validation

david-inglish:
## Description
Added validation code for input/ancillary files to check for existence, size and appropriate file extensions.

## Affected Issues
- #284 

## Testing
- Added tests to exercise the DISP-S1 input validation routines.


src/opera/util/input_validation.py
```diff
@@ -33,6 +33,8 @@ def check_input(input_object, logger, name, valid_extensions=None):
     valid_extensions : iterable, optional
         Expected file extensions of the file being validated. If not provided,
         no extension checking will take place.
+    check_zero_size : boolean, optional
+        If true, raise an exception for zero-size input objects
```
JimHofman, line 37: Might be a little clearer to say something like:  "If true, test for zero-size input."

JimHofman (approved):
Looks good.

src/opera/test/util/test_input_validation.py
```diff
@@ -0,0 +1,219 @@
...
+
+    def tearDown(self) -> None:
+        """
+        Return to starting directory
+        -------
+        """
+        os.chdir(self.test_dir)
+        self.working_dir.cleanup()
+
+    def test_validate_disp_inputs(self):
```
collinss-jpl, line 75: I appreciate the creation of the `test_input_validation.py` module, but I would prefer to see this test included in with the `test_disp_s1_pge.py` module.

src/opera/test/data/test_disp_input/compressed_slc_t087_185678_iw2_20180101_20180210.h5
```diff
@@ -0,0 +1 @@
+non-empty file
```
collinss-jpl, line 1: Rather than have these test files permanently checked in, I would rather they be dynamically created/removed in the `setUp` and `tearDown` logic of the unit test module, similar to what we do for the other PGEs.

collinss-jpl (changes requested):
The implementation itself looks great, I just had a few requests regarding how the unit tests and test data is organized.

collinss-jpl (approved):

PR #293, opened 2023-06-01
HLS Bounding Box Anti-meridian fix

### LINKED ISSUE #291: [Bug]:  Metadata bounding box not handling the antimeridian case
### Checked for duplicates

No - I haven't checked

### Describe the bug

When looking at Earthdata Search, tiles that span the antimeridian have bounding boxes that span the globe.

I believe the function causing the issue is `get_geographic_boundaries_from_mgrs_tile` (from https://github.com/nasa/opera-sds-pge/blob/33df7d19e940d9f4799438a4e4eeb010c0ce2abd/src/opera/util/metadata_utils.py#L120).
When not overlapping the antimeridian, it's doing the right thing.  However, when the tile overlaps the antimeridian, the western-most coordinate (in a local sense) is the one "to the left" of the antimeridian - i.e. the positive longitude value.

The function should be able to pass this unit test:
```
def test_get_geographic_boundaries_from_mgrs_tile_nominal_antimeridian(self):
    """Reproduce ADT results from values provided with code"""
    lat_min, lat_max, lon_min, lon_max = get_geographic_boundaries_from_mgrs_tile('60VXQ')

    self.assertAlmostEqual(lat_min, 62.173159300917874)
    self.assertAlmostEqual(lat_max, 63.20339285804396)
    self.assertAlmostEqual(lon_min, 178.98831533072024)
    self.assertAlmostEqual(lon_max, -178.96968891381724)
```
Currently, the function will output the same numbers as above, but `lon_min` and `lon_max` will be switched.

### What did you expect?

I expected the bounding boxes to correctly outline the shape of the tile, and be able to handle the case of straddling the antimeridian.

### Reproducible steps

```bash
This uses the function `get_geographic_boundaries_from_mgrs_tile` (from https://github.com/nasa/opera-sds-pge/blob/33df7d19e940d9f4799438a4e4eeb010c0ce2abd/src/opera/util/metadata_utils.py#L120)

1. Run `get_geographic_boundaries_from_mgrs_tile('T60VXQ')`
2. The function will report `lon_min = -178.96968891381724`, and `lon_max = 178.98831533072024`
3. For this case, the `lon_min` and `lon_max` should switch, due to crossing the antimeridian.
```

### Environment

_No response_

collinss-jpl:
## Description
- This branch addresses an issue in the`metadata_utils.get_geographic_boundaries_from_mgrs_tile` function for tiles that crossed the anti-meridian. The function has been updated with fixes that were applied to the PCM version of the function, which were originally provided by Gustavo Shiroma.

## Affected Issues
- Fixes #291 

## Testing
- The following unit test has been added to validate that the anti-meridian case is now handled correctly:

```python
    def test_get_geographic_boundaries_from_mgrs_tile_nominal_antimeridian(self):
        """Test MGRS tile code conversion with a tile that crosses the anti-meridian"""
        lat_min, lat_max, lon_min, lon_max = get_geographic_boundaries_from_mgrs_tile('T60VXQ')

        self.assertAlmostEqual(lat_min, 62.13198085489144)
        self.assertAlmostEqual(lat_max, 63.16076767648831)
        self.assertAlmostEqual(lon_min, 178.82637550795243)
        self.assertAlmostEqual(lon_max, -179.06925552951074)
```

Note that the lat/lon values differ slightly from the test case provided in the original bug report. This is because as part of this fix, 4.9km of margin is now added to the bounding box footprint to bring it in line with the HLS standard.

### DIFF
--- a/src/opera/test/util/test_metadata_utils.py
+++ b/src/opera/test/util/test_metadata_utils.py
@@ -42,25 +42,35 @@ class MetadataUtilsTestCase(unittest.TestCase):
         """Reproduce ADT results from values provided with code"""
         lat_min, lat_max, lon_min, lon_max = get_geographic_boundaries_from_mgrs_tile('15SXR')
 
-        self.assertAlmostEqual(lat_min, 31.616027943130398)
-        self.assertAlmostEqual(lat_max, 32.6212369766609)
-        self.assertAlmostEqual(lon_min, -91.94552881416524)
-        self.assertAlmostEqual(lon_max, -90.76425651871281)
+        self.assertAlmostEqual(lat_min, 31.572733739486036)
+        self.assertAlmostEqual(lat_max, 32.577473659397235)
+        self.assertAlmostEqual(lon_min, -91.99766472766642)
+        self.assertAlmostEqual(lon_max, -90.81751155385777)
 
     @skipIf(not osr_is_available(), reason="osgeo.osr is not installed on the local instance")
     def test_get_geographic_boundaries_from_mgrs_tile_leading_T(self):
         """Test MGRS tile code conversion when code starts with T"""
         lat_min, lat_max, lon_min, lon_max = get_geographic_boundaries_from_mgrs_tile('T15SXR')
 
-        self.assertAlmostEqual(lat_min, 31.616027943130398)
-        self.assertAlmostEqual(lat_max, 32.6212369766609)
-        self.assertAlmostEqual(lon_min, -91.94552881416524)
-        self.assertAlmostEqual(lon_max, -90.76425651871281)
+        self.assertAlmostEqual(lat_min, 31.572733739486036)
+        self.assertAlmostEqual(lat_max, 32.577473659397235)
+        self.assertAlmostEqual(lon_min, -91.99766472766642)
+        self.assertAlmostEqual(lon_max, -90.81751155385777)
 
     @skipIf(not osr_is_available(), reason="osgeo.osr is not installed on the local instance")
     def test_get_geographic_boundaries_from_mgrs_tile_invalid_tile(self):
         """Test MGRS tile code conversion with an invalid code"""
         self.assertRaises(RuntimeError, get_geographic_boundaries_from_mgrs_tile, 'X15SXR')
+
+    @skipIf(not osr_is_available(), reason="osgeo.osr is not installed on the local instance")
+    def test_get_geographic_boundaries_from_mgrs_tile_nominal_antimeridian(self):
+        """Test MGRS tile code conversion with a tile that crosses the anti-meridian"""
+        lat_min, lat_max, lon_min, lon_max = get_geographic_boundaries_from_mgrs_tile('T60VXQ')
+
+        self.assertAlmostEqual(lat_min, 62.13198085489144)
+        self.assertAlmostEqual(lat_max, 63.16076767648831)
+        self.assertAlmostEqual(lon_min, 178.82637550795243)
+        self.assertAlmostEqual(lon_max, -179.06925552951074)
 
     def test_get_rtc_s1_product_metadata(self):
         """Test retrieval of product metadata from HDF5 files"""
--- a/src/opera/test/util/test_metadata_utils.py
+++ b/src/opera/test/util/test_metadata_utils.py
@@ -70,7 +70,7 @@ class MetadataUtilsTestCase(unittest.TestCase):
         self.assertAlmostEqual(lat_min, 62.13198085489144)
         self.assertAlmostEqual(lat_max, 63.16076767648831)
         self.assertAlmostEqual(lon_min, 178.82637550795243)
-        self.assertAlmostEqual(lon_max, -179.06925552951074)
+        self.assertAlmostEqual(lon_max, -178.93677941363356)
 
     def test_get_rtc_s1_product_metadata(self):
         """Test retrieval of product metadata from HDF5 files"""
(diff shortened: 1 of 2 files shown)


src/opera/util/metadata_utils.py
```diff
@@ -185,17 +186,50 @@ def get_geographic_boundaries_from_mgrs_tile(mgrs_tile_name):
...
+            # values are both at the same side of the dateline (either left
+            # or right).
+            #
+            # The conditions `> 100` and `< 100` are used to test if the
+            # longitude point is on the left side of the antimeridian crossing
+            # (`> 100`) or at the right side (`< 100`)
+            #
+            # We also want to check if the point is at the west or east side of the tile.
+            # Points at the west, i.e, where offset_x_multiplier == 0 may update `lon_min`
+            if (offset_x_multiplier == 0 and
```
sjlewis-jpl, line 221: I think the if-statements for `lon_min` and `lon_max` could be simplified.  Add `offset_x_multiplier ==0 and ...` on the previous if-statement.  Something like this:
```
if (offset_x_multiplier == 0) and (lon_min is None or lon_min > lon):
    lon_min = lon
if (offset_x_multiplier == 1) and (lon_max is None or lon_max < lon):
    lon_max = lon
``` 

That should pick the `lon_min` from the two boundary points on the "left-hand" side of the granule (viewing when north is up), and `lon_max` from the two boundary points on the "right-hand" side.

src/opera/test/util/test_metadata_utils.py
```diff
@@ -42,26 +42,36 @@ def test_get_geographic_boundaries_from_mgrs_tile_nominal(self):
...
         lat_min, lat_max, lon_min, lon_max = get_geographic_boundaries_from_mgrs_tile('T15SXR')
 
-        self.assertAlmostEqual(lat_min, 31.616027943130398)
-        self.assertAlmostEqual(lat_max, 32.6212369766609)
-        self.assertAlmostEqual(lon_min, -91.94552881416524)
-        self.assertAlmostEqual(lon_max, -90.76425651871281)
+        self.assertAlmostEqual(lat_min, 31.572733739486036)
+        self.assertAlmostEqual(lat_max, 32.577473659397235)
+        self.assertAlmostEqual(lon_min, -91.99766472766642)
+        self.assertAlmostEqual(lon_max, -90.81751155385777)
```
sjlewis-jpl, line 58: I see why these changed (from lines 191 and 192 in the new function).  But it is a bit concerning that our lat/lon values seem to differ from the bounding polygons from the Sentinel-2 UTM Tiling Grid (HLS is supposed to be using the exact same tiling grid).  You can find a kml of that tiling grid here:  https://eatlas.org.au/data/uuid/f7468d15-12be-4e3f-a246-b2882a324f59

Grepping through that file for the tile IDs of the ones used in the unit tests (my grep commands copied below) shows slightly different lat/lon bounds.  I assume this is due to us using a slightly different projection for our calculations.  If these outputs are only used for showing bounding boxes in Earthdata Search, say, then probably not a big deal.  But worth a conversation with ADT.
```
grep -A 20 "15SXR" S2A_OPER_GIP_TILPAR_MPC__20151209T095117_V20150622T000000_21000101T000000_B00.kml
grep -A 20 "60VXQ" S2A_OPER_GIP_TILPAR_MPC__20151209T095117_V20150622T000000_21000101T000000_B00.kml
```

sjlewis-jpl (changes requested):
The change that I mentioned to the if-statements should be applied (https://github.com/nasa/opera-sds-pge/pull/293/files#r1218582327).  As is, the function gives the right answer for `lon_min`, but not for `lon_max`.

If you insert this line just after `transformation.TransformPoint` is called, you can see the lat/lon pairs of each corner:
```
print(offset_x_multiplier, offset_y_multiplier, x, y, lat, lon, z)
```
Then running the function shows this:
```
>>> get_geographic_boundaries_from_mgrs_tile('T60VXQ')
0 0 595100.0 6895100.0 62.17575213937371 178.82637550795243 0.0
0 1 595100.0 7004900.0 63.16076767648831 178.88808853630766 0.0
1 0 704900.0 6895100.0 62.13198085489144 -179.06925552951074 0.0
1 1 704900.0 7004900.0 63.11512337751825 -178.93677941363356 0.0
Out[25]: (62.13198085489144, 63.16076767648831, 178.82637550795243, -179.06925552951074)
```
From the printed lines, the furthest east point should be the largest of the two negative longitudes (i.e. the "least negative"), `-178.93677941363356`.  But it's giving the other as the `lon_max`.

If you swap out the if-statements for the ones in my earlier comment, you should see this:
```
>>> get_geographic_boundaries_from_mgrs_tile_new('T60VXQ')
0 0 595100.0 6895100.0 62.17575213937371 178.82637550795243 0.0
0 1 595100.0 7004900.0 63.16076767648831 178.88808853630766 0.0
1 0 704900.0 6895100.0 62.13198085489144 -179.06925552951074 0.0
1 1 704900.0 7004900.0 63.11512337751825 -178.93677941363356 0.0
Out[35]: (62.13198085489144, 63.16076767648831, 178.82637550795243, -178.93677941363356)
```
That one pulled the correct `lon_max`.

collinss-jpl:
@sjlewis-jpl I've updated the conditionals for lon_max/lon_min according to your feedback, and ensured that the numbers I get match what you've posted. Let me know if anything else still looks off.

sjlewis-jpl (approved):

Issue #298, opened 2023-06-12
DISP-S1 Integration Test Pipeline


collinss-jpl:
Ticket to track setup of Integration Test pipeline for DISP-S1 PGE. This includes formation and staging of the expected input/output datasets in S3. Note for DISP-S1, the integration test should utilize the "minimal" set of expected inputs and outputs referenced in #283 .

collinss-jpl:
Resolved by #308

PR #308, opened 2023-06-26
298 DISP-S1 Integration Test Pipeline data staging and product comparison

david-inglish:
## Description
- These changes add the script to test the integration of DISP-S1 and add the test data to be used.
- The PGE was updated to be able to process the test data.
- Water mask is referred to as an array in a comment, but is a string in the SAS yamale validation.  It has been switched to be a string.

## Affected Issues
- #298 

## Testing
- The Build/Integration pipeline is unsuccessful because the PGE output does not match the golden test data.  This is the expected result until updated golden test data is available.


.ci/scripts/test_int_disp_s1.sh
```diff
@@ -0,0 +1,142 @@
...
+    echo "docker exit indicates failure: ${docker_exit_status}"
+    overall_status=1
+else
+    echo "<tr><th>Compare Result</th><th><ul><li>Expected file</li><li>Output file</li></ul></th><th>disp_validate_product_opera_pge.py output</th></tr>" >> "$RESULTS_FILE"
+
+    output_file="20180101_20180330.unw.nc"
+    expected_file="20180101_20180330.unw.nc"
+
+    docker_out=$(docker run --rm \
+						    -v "${output_dir}":/out:ro \
```
collinss-jpl, line 114: Looks like some tabs snuck in that need to be converted to spaces

collinss-jpl (commented):
Looks good, but remind me why this script isn't running the PGE yet? I thought what we have currently should be able to run and produce results.

collinss-jpl (approved):
Looks great, just the small merge conflict that needs to be resolved and this is ready to go.

PR #318, opened 2023-07-10
304 dswx s1 mock gdal

JimHofman:
## Description
Set up Mock GDAL requests for DSWX-S1.  This code is shared with DSWX-HLS so verify all DSWX-HLS unit tests complete successfully.

## Affected Issues
N/A

## Testing
Verify all unit tests run successfully


collinss-jpl (approved):
Looks great @JimHofman, and thanks for the docstring cleanup!

PR #320, opened 2023-07-18
317 jenkins clean up

JimHofman:
## Description
Add Jenkins file to document simple pipeline to clean up temporary files

## Affected Issues
n/a

## Testing
This pipeline did what it was designed to do

### DIFF
--- a/.ci/jenkins/delete-tmp-dirs/Jenkinsfile
+++ b/.ci/jenkins/delete-tmp-dirs/Jenkinsfile
@@ -1,0 +1,13 @@
+#!/usr/bin/env groovy
+
+pipeline {
+    agent any
+    stages {
+        stage("clean_up")  {
+            steps {
+                echo "Cleaning up data/tmp directories"
+                rm -rf tmp.*
+            }
+        }
+    }
+]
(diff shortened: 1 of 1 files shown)


.ci/jenkins/delete-tmp-dirs/Jenkinsfile
```diff
@@ -0,0 +1,19 @@
...
+disc space in the add/tmp directory on the opera-pge-ci aws machine.
+Using a Jenkins pipeline was required due to security concerns.
+*/
+pipeline {
+    agent any
+    stages {
+        stage("clean_up")  {
+            steps {
+                echo "Cleaning up data/tmp directories"
+                sh 'rm -rf tmp.*'
```
collinss-jpl, line 15: This does not match the deployed version, which has this line set to `sh 'rm -rf /data/tmp/tmp.*'`. I'd prefer the two version be kept in sync as much as possible.

collinss-jpl (changes requested):
Looks good, just need to update this version to match what is currently configured within the actual pipeline.

collinss-jpl (approved):

PR #341, opened 2023-08-16
# 309 Burst ID validation for DISP-S1 inputs

JimHofman:
## Description
- # 309 Burst ID validation for DISP-S1 inputs

## Affected Issues
- N/A

## Testing
- unit tests cover 100% of the new code

### DIFF
--- a/src/opera/test/pge/disp_s1/test_disp_s1_pge.py
+++ b/src/opera/test/pge/disp_s1/test_disp_s1_pge.py
@@ -198,7 +198,7 @@ class DispS1PgeTestCase(unittest.TestCase):
             runconfig_dict = yaml.safe_load(infile)
 
         runconfig_dict['RunConfig']['Groups']['PGE']['PrimaryExecutable']['AlgorithmParametersSchemaPath'] = \
-                       'test/data/test_algorithm_parameters_non_existent.yaml'  # noqa E211
+            'test/data/test_algorithm_parameters_non_existent.yaml'  # noqa E211
 
         with open(test_runconfig_path, 'w', encoding='utf-8') as outfile:
             yaml.safe_dump(runconfig_dict, outfile, sort_keys=False)
@@ -249,8 +249,8 @@ class DispS1PgeTestCase(unittest.TestCase):
         with open(runconfig_path, 'r', encoding='utf-8') as infile:
             runconfig_dict = yaml.safe_load(infile)
 
-        runconfig_dict['RunConfig']['Groups']['SAS']['dynamic_ancillary_file_group']\
-                      ['algorithm_parameters_file'] = 'test/data/test_algorithm_parameters_non_existent.yaml' # noqa E211
+        runconfig_dict['RunConfig']['Groups']['SAS']['dynamic_ancillary_file_group'] \
+            ['algorithm_parameters_file'] = 'test/data/test_algorithm_parameters_non_existent.yaml'  # noqa E211
 
         with open(test_runconfig_path, 'w', encoding='utf-8') as outfile:
             yaml.safe_dump(runconfig_dict, outfile, sort_keys=False)
@@ -275,12 +275,12 @@ class DispS1PgeTestCase(unittest.TestCase):
         # Test non-existent file detection
         test_filename = 't087_123456_iw2_non_existent_input_file'
         sas_config = {
-            'input_file_group' : {
-                'cslc_file_list' : [
+            'input_file_group': {
+                'cslc_file_list': [
                     test_filename
                 ]
             },
-            'dynamic_ancillary_file_group' : {
+            'dynamic_ancillary_file_group': {
             }
         }
         runconfig = MockRunConfig(sas_config)
@@ -397,12 +397,13 @@ class DispS1PgeTestCase(unittest.TestCase):
         self.assertIn('overall.log_messages.critical: 0', log)
 
         files_to_remove = (cslc_file_list + amplitude_dispersion_files +
-                          amplitude_mean_files + geometry_files +
-                          tec_files + weather_model_files)
+                           amplitude_mean_files + geometry_files +
+                           tec_files + weather_model_files)
         files_to_remove.append(mask_file)
         files_to_remove.append(dem_file)
         for f in files_to_remove:
             os.remove(f)
+
 
     def test_get_cslc_input_burst_id_set(self):
         """
@@ -410,6 +411,25 @@ class DispS1PgeTestCase(unittest.TestCase):
         cslc_input_files, amplitude_dispersion_files, amplitude_mean_files,
         and geometry_files.
         """
+
+        def get_sample_input_files(file_type: str) -> list:
+            """Helper function for test_get_cslc_input_burst_id_set()"""
+            if file_type == 'compressed':
+                return ['compressed_slc_t087_185683_iw2_220180101_20180210.h5',
+                        'compressed_slc_t087_185684_iw2_20180101_20180210.h5']
+            elif file_type == 'uncompressed':
+                return ['t087_185683_iw2_20180222_VV.h5', 't087_185683_iw2_20180306_VV.h5',
+                        't087_185684_iw2_20180222_VV.h5', 't087_185684_iw2_20180306_VV.h5']
+            else:
+                return []
+
+        def add_text_to_file(file_list: list) -> list:
+            """Helper function to add a bit of text to each empty file in a list"""
+            for f in file_list:
+                with open(f, 'w') as wf:
+                    wf.write('Hello\n')
+            return file_list
+
         sas_config = {
             'input_file_group': {
                 'cslc_file_list': [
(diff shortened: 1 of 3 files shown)


src/opera/test/pge/disp_s1/test_disp_s1_pge.py
```diff
@@ -404,6 +404,158 @@ def test_disp_s1_pge_validate_inputs(self):
         for f in files_to_remove:
             os.remove(f)
 
+    def test_get_cslc_input_burst_id_set(self):
+        """
+        Set of tests to sanity check the burst ideas associated with:
```
david-inglish, line 409: should this be "burst ids" instead of "burst ideas"?

src/opera/test/pge/disp_s1/test_disp_s1_pge.py
```diff
@@ -404,6 +404,158 @@ def test_disp_s1_pge_validate_inputs(self):
...
+            validate_disp_inputs(runconfig, logger, "DISP-S1")
+        # Check to see that the RuntimeError is as expected
+        logger.close_log_stream()
+        log_file = logger.get_file_name()
+        self.assertTrue(exists(log_file))
+        with open(log_file, 'r', encoding='utf-8') as lfile:
+            log = lfile.read()
+        self.assertIn('input_burst_ids: ', log)
+        self.assertIn('do not match ancillary_burst_ids: ', log)
+        # Test for an ancillary file the does not have a unique burst id
```
david-inglish, line 530: change "the" to "that"

src/opera/util/input_validation.py
```diff
@@ -130,6 +131,133 @@ def validate_slc_s1_inputs(runconfig, logger, name):
...
+
+    compressed_ids = get_burst_id_set(compressed_input_file_list, logger, name)
+    single_file_burst_id_set = get_burst_id_set(single_input_file_list, logger, name)
+
+    # Case 1:  uncompressed files only in cslc inputs
+    if len(compressed_ids) == 0:
+        return single_file_burst_id_set
+    # Case 2: uncompressed files and compressed files in cslc inputs with non-matching burst ids
+    elif single_file_burst_id_set != compressed_ids:
+        msg = f"single_file_burst_id_set: {single_file_burst_id_set} does not match compressed_file_burst_id_set: " \
```
david-inglish, line 253: should "compressed_file_burst_id_set"  be "compressed_ids" (or perhaps both be "compressed_file_burst_id_set") ?

david-inglish (approved):
Looks good.  I've noted some nitpick things here and there.

src/opera/test/pge/disp_s1/test_disp_s1_pge.py
JimHofman: haha!  oh brother, I can't believe I let that one by.

src/opera/util/input_validation.py
JimHofman: I'll change compressed_ids to compressed_file_burst_id_set to make things more clear.  Thanks.

src/opera/util/input_validation.py
```diff
@@ -130,6 +131,133 @@ def validate_slc_s1_inputs(runconfig, logger, name):
...
+        if match:
+            burst_ids.add(re.findall(reg_ex, i)[0])
+        else:
+            msg = f'Input file present without properly formatted burst_id: {i}'
+            logger.critical(name, ErrorCode.INVALID_INPUT, msg)
+
+    return burst_ids
+
+
+def check_amp_mean_geometry_burst_ids(cslc_input_burst_ids, ancillary_file_list, logger, name):
```
collinss-jpl, line 171: I love that this function is written at the abstraction level of "ancillary files", such that it will work for any arbitrary list of input ancillaries that contain burst IDs. Because of this, I think the name of the function should be changed to match, something like `check_disp_s1_ancillary_burst_ids`

src/opera/util/input_validation.py
```diff
@@ -130,6 +131,133 @@ def validate_slc_s1_inputs(runconfig, logger, name):
...
+    ------
+    RuntimeError
+        If there are ancillary files of the same type with the same burst_id.
+        If the set of ancillary file burst_ids do not match the set of cslc input file burst_ids.
+
+    """
+    ancillary_burst_ids = get_burst_id_set(ancillary_file_list, logger, name)
+    # Test that there is a unique burst_id for each burst
+    if len(ancillary_burst_ids) != len(ancillary_file_list):
+        msg = f'Burst ids in {ancillary_file_list} are not unique.'
```
collinss-jpl, line 201: I'm not sure "uniqueness" is the proper way to phrase this error message, better to just state that "The number of burst ID's between CSLC and ancillaries do not match", and then also print the length of each set.

src/opera/util/input_validation.py
```diff
@@ -130,6 +131,133 @@ def validate_slc_s1_inputs(runconfig, logger, name):
...
+        logger.critical(name, ErrorCode.INVALID_INPUT, msg)
+    # Verify that the sets of burst_ids are equal
+    if ancillary_burst_ids != cslc_input_burst_ids:
+        msg = f'input_burst_ids: {cslc_input_burst_ids} do not match ancillary_burst_ids: ' \
```
collinss-jpl, line 205: The formatting of this error message might be hard to read in production where there could be up to 30 burst ids printed out from each set, so we might want to add some line breaks and only print out the differences between the two sets, like so:

```python
msg = ("Input CSLC burst IDs do not match the set of ancillary burst IDs\n"
        f"{cslc_input_burst_ids.difference(ancillary_burst_ids)=\n}
        f"{ancillary_burst_ids.difference(cslc_input_burst_ids)=\n}")
```

src/opera/util/input_validation.py
```diff
@@ -130,6 +131,133 @@ def validate_slc_s1_inputs(runconfig, logger, name):
...
+    Raises
+    ------
+    RuntimeError
+        If the compressed burst_ids set does not match the uncompressed burst_id set.
+
+    """
+    compressed_input_file_list = []
+    single_input_file_list = []
+
+    for i in cslc_input_file_list:
```
collinss-jpl, line 241: Lets compress this down to:

```python
compressed_input_file_list = list(filter(lambda filename: 'compressed' in filename), cslc_input_file_list)
single_input_file_list = list(set(cslc_input_file_list) - set(compressed_input_file_list))
```

src/opera/util/input_validation.py
collinss-jpl, line 255: See my message above about how to format this error message to only print out the differences between the two sets

src/opera/util/input_validation.py
```diff
@@ -130,6 +131,133 @@ def validate_slc_s1_inputs(runconfig, logger, name):
...
+    cslc_input_file_list : list
+        List of all cslc input files pulled from the runconfig file
+    logger : PgeLogger
+        Logger passed by the calling PGE
+    name : str
+        PGE name
+
+    Returns
+    -------
+    burst_id_set :set
```
collinss-jpl: Need a description for this return value

src/opera/test/pge/disp_s1/test_disp_s1_pge.py
```diff
@@ -404,6 +404,158 @@ def test_disp_s1_pge_validate_inputs(self):
...
+
+                ]
+            },
+            'dynamic_ancillary_file_group': {
+            }
+        }
+        # Test uncompressed files only as input
+        cslc_file_list = ['t087_185683_iw2_20180222_VV.h5', 't087_185683_iw2_20180306_VV.h5',
+                          't087_185684_iw2_20180222_VV.h5', 't087_185684_iw2_20180306_VV.h5']
+        for f in cslc_file_list:
```
collinss-jpl, line 425: This block from 425-427 is repeated enough that I think it merits its own helper function.

collinss-jpl (changes requested):
All of the logic looks great, but there are some improvements I'd like to see to the error messages before this gets merged.

src/opera/util/input_validation.py
```diff
@@ -195,15 +195,18 @@ def check_amp_mean_geometry_burst_ids(cslc_input_burst_ids, ancillary_file_list,
...
+        # msg = f'Burst ids in {ancillary_file_list} are not unique.'
         logger.critical(name, ErrorCode.INVALID_INPUT, msg)
-    # Verify that the sets of burst_ids are equal
+    # Verify that the sets of ancillary input burst ID's match the set of CSLC input burst ID's
     if ancillary_burst_ids != cslc_input_burst_ids:
-        msg = f'input_burst_ids: {cslc_input_burst_ids} do not match ancillary_burst_ids: ' \
-              f'{ancillary_burst_ids} for {ancillary_file_list}'
+        msg = f"Set of input CSLC burst IDs do not match the set of ancillary burst IDs: " \
+              f"In cslc set, but not in ancillary set: {cslc_input_burst_ids.difference(ancillary_burst_ids)} " \
+              f"In ancillary set, but not in cslc set: {ancillary_burst_ids.difference(cslc_input_burst_ids)}"
```
collinss-jpl, line 209: New message looks good, but as is it would print as one long line which might be hard to read. Let's add some "\n" line breaks to the end of each string so the formatting matches the way you've broken up the string here.

src/opera/util/input_validation.py
```diff
@@ -235,25 +238,24 @@ def get_cslc_input_burst_id_set(cslc_input_file_list, logger, name):
...
         return single_file_burst_id_set
     # Case 2: uncompressed files and compressed files in cslc inputs with non-matching burst ids
     elif single_file_burst_id_set != compressed_file_burst_id_set:
-        msg = f"single_file_burst_id_set: {single_file_burst_id_set} does not match compressed_file_burst_id_set: " \
-              f"'{compressed_file_burst_id_set}'"
+        msg = f"Set of input CSLC 'compressed' burst IDs do not match the set of 'uncompressed' burst IDs: " \
+              f"In 'compressed' set, but not in 'uncompressed' set: " \
+              f"{compressed_file_burst_id_set.difference(single_file_burst_id_set)} " \
+              f"In 'uncompressed' set, but not in 'compressed' set:" \
+              f" {single_file_burst_id_set.difference(compressed_file_burst_id_set)}"
```
collinss-jpl, line 258: Same as my comment above regarding line breaks

src/opera/util/input_validation.py
```diff
@@ -131,7 +131,7 @@ def validate_slc_s1_inputs(runconfig, logger, name):
             logger.critical(name, ErrorCode.INVALID_INPUT, error_msg)
 
 
-def get_burst_id_set(input_file_group, logger, name):
+def check_disp_s1_ancillary_burst_ids(input_file_group, logger, name):
```
collinss-jpl, line 134: My intent was not to have this function renamed, but the one below it (check_amp
_mean_geometry_burst_ids). As-is, this function now has a misleading name (is not specific to ancillaries), and the one below is still too specific (could be applied to any hypothetical ancillary group, not just the amp/mean/geometry files).

collinss-jpl (changes requested):
Getting closer, another round of fixes and this branch should be good to merge.

src/opera/util/input_validation.py
JimHofman: Makes sense haha.

collinss-jpl (approved):

Issue #339, opened 2023-08-15
DISP-S1 product validation tool should have a filter for datasets to ignore during validation 


david-inglish:
### Checked for duplicates

Yes - I've already checked

### Alternatives considered

Yes - and alternatives don't suffice

### Describe the feature request

The DISP-S1 validation tool checks all datasets within the product files being compared.  We need the validation tool to be configurable to ignore some datasets, specifically the contents of the runconfig in the DISP-S1 product since the paths to filenames can be different between the delivered SAS runconfig and the PGE runconfig.

collinss-jpl:
Addressed in #344

Issue #326, opened 2023-07-31
Re-enable PNG output product validation for DISP-S1


collinss-jpl:

collinss-jpl:
Addressed in #344

PR #344, opened 2023-08-18
326 Re-enable png output product validation for DISP-S1

david-inglish:
## Description
- Code that was written to validate the .png product needed to be enabled and tested with an updated DISP-S1 SAS delivery.
- A change was made to the DISP-S1 validation tool to ignore selective groups during comparison, and the pge_runconfig is now being excluded from comparison during the integration test.

## Affected Issues
- #326 
- #339 

## Testing
- Testing was done at the unit and integration level.

### DIFF
--- a/.ci/scripts/disp_validate_product_opera_pge.py
+++ b/.ci/scripts/disp_validate_product_opera_pge.py
@@ -62,6 +62,7 @@ def compare_groups(
     test_group: h5py.Group,
     pixels_failed_threshold: float = 0.01,
     diff_threshold: float = 1e-5,
+    exclude_groups: list = []
 ) -> None:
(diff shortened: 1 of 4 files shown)


.ci/scripts/disp_validate_product_opera_pge.py
```diff
@@ -62,6 +62,7 @@ def compare_groups(
     test_group: h5py.Group,
     pixels_failed_threshold: float = 0.01,
     diff_threshold: float = 1e-5,
+    exclude_groups: list = []
```
collinss-jpl, line 65: Docstring below should be updated to describe use of this arg

collinss-jpl (approved):
Looks good! Only had one minor finding, but this can be merged whenever.

PR #349, opened 2023-08-21
DISP-S1 ISO XML Template Addition

### LINKED ISSUE #329: DISP-S1 ISO XML Metadata Template
Ticket to track creation of the jinja2 template for the DISP-S1 ISO XML metadata.

collinss-jpl:
## Description
- This branch adds the initial jinja2 template for the DISP-S1 ISO XML metadata.

## Affected Issues
- Resolves #329 

## Testing
- No executable code added in this branch, so no affect to existing tests

### DIFF
--- a/src/opera/pge/disp_s1/templates/OPERA_ISO_metadata_L3_DISP_S1_template.xml.jinja2
+++ b/src/opera/pge/disp_s1/templates/OPERA_ISO_metadata_L3_DISP_S1_template.xml.jinja2
@@ -1,0 +1,900 @@
+<?xml version='1.0' encoding='utf-8'?>
+<gmi:MI_Metadata xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                 xmlns:gmd="http://www.isotc211.org/2005/gmd"
+                 xmlns:gco="http://www.isotc211.org/2005/gco"
+                 xmlns:eos="http://earthdata.nasa.gov/schema/eos"
+                 xmlns:xlink="http://www.w3.org/1999/xlink"
+                 xmlns:gml="http://www.opengis.net/gml/3.2"
+                 xmlns:gmi="http://www.isotc211.org/2005/gmi"
+                 xmlns:gmx="http://www.isotc211.org/2005/gmx">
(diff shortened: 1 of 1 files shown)


src/opera/pge/disp_s1/templates/OPERA_ISO_metadata_L3_DISP_S1_template.xml.jinja2
```diff
@@ -0,0 +1,900 @@
...
+            </gmd:referenceSystemIdentifier>
+        </gmd:MD_ReferenceSystem>
+    </gmd:referenceSystemInfo>
+    <!-- This section documents granule data in this metadata record. -->
+    <gmd:identificationInfo>
+        <gmd:MD_DataIdentification>
+            <!-- This section holds the granule MetadataProviderDates, and the granule identifiers -->
+            <gmd:citation>
+                <gmd:CI_Citation>
+                    <gmd:title>
```
JimHofman, line 138: Just looking at the pattern it looks as if the comment:   might be needed.

JimHofman, line 138: Question:  Do lines such at this with '{# <name>}<' in the line imply "this will be added by the PGE"?  If so then for consistency we should add the comment.  There are a few more like this below if this comment has any meaning.

JimHofman (approved):
Looked things over.  I have a syntax question. What is the result when a variable in parenthesis is lead with '#', e.g. 
<gco:CharacterString>{{ custom_data.ISO_OPERA_FilePackageName }}_**{#_** _**ISO_OPERA_FilePackageName #}**_</gco:CharacterString>

collinss-jpl:
> Looked things over. I have a syntax question. What is the result of a variable name being bracketed by '{# #}'?

As it turns out, it just means the block is considered a comment by the template renderer.

I had though that the {# #} blocks might provide a name for field that would show up in logging statements to help troubleshoot rendering issues. The use of them were holdovers from the examples I got from SWOT, so I thought they might be useful or necessary, but this does not seem to be the case.

Issue #342, opened 2023-08-17
[New Feature]: Blacklist/ignore list for HD5 metadata extraction utility


collinss-jpl:
### Checked for duplicates

Yes - I've already checked

### Alternatives considered

Yes - and alternatives don't suffice

### Related problems

The current implementation of `metadata_utils.convert_h5py_group_to_dict` will always convert every available key/value pair under the provided group object, but for certain groupings this can include large numpy arrays that we don't actually care about, causing memory over-consumption and significant read times while we extract from multiple output products.

### Describe the feature request

`convert_h5py_group_to_dict` should be updated with an optional argument that specifies a blacklist of keys that should be ignored during traversal. This would allow us to avoid reading large raster layers which have to be purged from memory as soon as they are read in.

PR #350, opened 2023-08-24
342 Ignore list for hdf5 metadata extraction utility

david-inglish:
## Description
- These changes implement a feature to exclude group Datasets from HDF5 files when retrieving metadata.  This will save on memory use where large Datasets are present in the group.
- CSLC-S1 has been updated to use the new feature.

## Affected Issues
- #342 

## Testing
- pytest assertions have been added to check that datasets have been ignored.


JimHofman (approved):
Looks good

src/opera/util/metadata_utils.py
```diff
@@ -476,9 +481,18 @@ def get_cslc_s1_product_metadata(file_name):
         python dict containing the HDF5 file metadata which is used in the
         ISO template.
     """
+    # Remove some of the larger arrays from the metadata, so we don't use
```
collinss-jpl, line 484: Lets change "Remove" to "Ignore" here

collinss-jpl (changes requested):
Very minor changes requested, but ones I'd like to see before this is merged

collinss-jpl (approved):

Issue #347, opened 2023-08-21
DISP-S1 File Naming Convention Application


collinss-jpl:
Ticket to track implementation of file naming convention application for the DISP-S1 PGE. According to the latest (interface) product spec for DISP-S1, the naming conventions are as follows:

```
OPERA_L3_DISP-S1_<Mode>_<FrameID>_<Polarization>_<ReferenceDateTime>_<SecondaryDateTime>_<ProductVersion>_<ProductGenerationDateTime>.nc

where:

- Mode: S1-A/B acquisition mode 
  * Should be fixed to "IW"
- FrameID: Unique frame identification number as a 5-digit string in the format FXXXXX
- Polarization:  polarization of the input bursts 
    * e.g., VV or VH, this should be derivable from one of the names of the input CSLCs in the RunConfig
- ReferenceDateTime: The acquisition sensing start date and time of the input satellite imagery for the first burst in the frame of the reference product in the format YYYYMMDDTHHMMSSZ 
    * This should come from the product metadata under "identification/reference_datetime", however, it is missing from the current delivery. We should assign a placeholder datetime in the meantime
- SecondaryDateTime: The acquisition sensing start date and time of the input satellite imagery for the first burst in the frame of this secondary product in the format YYYYMMDDTHHMMSSZ
    * Similar to ReferenceDateTime, in that it should be derived from "identification/secondary_datetime", but we will need a placeholder for now
- ProductVersion: OPERA DISP-S1 product version number with four characters, including the letter “v” and two digits indicating the major and minor versions, which are delimited by a period
    * Available from product metadata
- ProductGenerationDateTime: The date and time at which the product was generated by OPERA with the format of YYYYMMDDTHHMMSSZ
   * get_time_for_filename(self.production_datetime) will return the appropriate value

An example of a DISP-S1 product file name:
 
OPERA_DISP-S1_IW_F01234_VV_20190101T232711Z_20190906T232711Z_v1.0_20230101T100506Z.nc

```

Besides the .nc output product, the output .png product should also be renamed using the same conventions, but just with the .png extension.

The new file renaming functions should also be configured within the `rename_by_pattern_map` defined in the `DispS1Executor` class. For example:

```
self.rename_by_pattern_map = {
    "*.nc": self._netcdf_filename,
    "*.png": self._png_filename
}
```

In addition to adding the naming conventions, the unit tests for DISP-S1 should be updated accordingly to test their application.

[...]

PR #361, opened 2023-09-11
347 DISP-S1 File Naming Convention Application

david-inglish:
## Description
- This is the implementation for the file naming convention for DISP-S1 products

## Affected Issues
- #347 
- 
## Testing
- A unit test has been updated to included a test that validates the file naming convention.


src/opera/pge/disp_s1/disp_s1_pge.py
```diff
@@ -139,6 +142,188 @@ def _validate_output(self):
...
+        mode = "IW"
+
+        # FrameID: Unique frame identification number as a 5-digit string in
+        # the format FXXXXX
+        frame_id = f"F{disp_metadata['identification']['frame_id']:05d}"
+
+        # Polarization:  polarization of the input bursts
+        # derived from names of input CSLCs in the runconfig
+        cslc_file_list = self.runconfig.sas_config['input_file_group']['cslc_file_list']
+        pattern = re.compile(r"t\w{3}_\d{6}_iw[1|2|3]_[0-9]{8}_(VV|VH).h5")
```
collinss-jpl, line 234: To ensure we've covered all bases, the list of possible polarization values should be `VV|VH|HH|HV|VV\+VH|HH\+HV`

src/opera/pge/disp_s1/disp_s1_pge.py
```diff
@@ -139,6 +142,188 @@ def _validate_output(self):
...
+        for cslc_file in cslc_file_list:
+            cslc_file_basename = basename(cslc_file)
+            if pattern.match(cslc_file_basename):
+                polarization = cslc_file_basename.split('_')[-1].split('.')[0]
```
collinss-jpl, line 238: We can simplify this a bit by leveraging group-matching in the regular expression:

If the regex is modified to `r"t\w{3}_\d{6}_iw[1|2|3]_[0-9]{8}_(?<polarization>VV|VH).h5"`, we can then extract `polarization` from the match object group dictionary:

```
result = pattern.match(cslc_file_basename):
if result:
    polarization = result.groupdict()['polarization']
```

src/opera/pge/disp_s1/disp_s1_pge.py
```diff
@@ -139,6 +142,188 @@ def _validate_output(self):
...
+        # "identification/reference_datetime", however, it is missing from
+        # the current delivery. We should assign a placeholder datetime in
+        # the meantime
+        reference_date_time = "YYYYMMDDTHHMMSSZ"
+
+        # SecondaryDateTime: The acquisition sensing start date and time of
+        # the input satellite imagery for the first burst in the frame of this
+        # secondary product in the format YYYYMMDDTHHMMSSZ
+        # TODO This should come from the product metadata
+        secondary_date_time = "YYYYMMDDTHHMMSSZ"
```
collinss-jpl, line 259: Rather than define the placeholders here, I would prefer to see them in `_collect_disp_s1_product_metadata`, since that has been the traditional place where the PGE makes modifications to the metadata dictionary to make up for shortfalls in the current product metadata.

Also, rather than the literal "YYYYMMDDTHHMMSSZ", we should use an actual datetime string since PCM will be expecting an actual datetime to pattern match against. The actual dates themselves can be completely made up so its obvious they are placeholders, such as "20230101T000000Z"

src/opera/pge/disp_s1/disp_s1_pge.py
```diff
@@ -139,6 +142,188 @@ def _validate_output(self):
...
+        return self._cached_core_filename
+
+    def _frame_filename(self, inter_filename=None):
+        """
+        Returns the file name to use for frame-based DISP products produced
+        by this PGE.
+
+        The filename for the DISP-S1 frame products consists of:
+
+        <CoreFilename>_<Mode>_<FrameID>_<Polarization>_<ReferenceDateTime>_<SecondaryDateTime>_<ProductVersion>_<ProductGenerationDateTime>
```
collinss-jpl, line 198: Spacing looks a bit off here? Should be aligned with the line above it.

collinss-jpl (changes requested):
Implementation looks great, just had a few minor corrections before this is ready to be merged. Great work @david-inglish!

JimHofman (approved):
Nothing to add.

collinss-jpl (approved):
Changes look great, thanks as always @david-inglish!

PR #366, opened 2023-09-13
CSLC-S1 SAS Final PGE Integration

### LINKED ISSUE #352: CSLC-S1 SAS Final PGE Integration
Ticket to track integration of the CSLC-S1 Final SAS delivery with the PGE library. 

In addition to the basic reintegration steps, this ticket should also include any updates needed to the ISO XML template to make sure all appropriate metadata is being collected, and that the integration test pipeline is updated with the latest assets for the Final delivery.

collinss-jpl:
## Description
- This branch integrates the Final SAS delivery for CSLC-S1 with the PGE library. Expected assets for the integration test have been updated, as well as the example RunConfig. This branch also updates the base PGE schema to allow product version to be a string or a number, so we can support specifying multi-part version numbers (i.e. 1.0.0 and not just 1.0).

## Affected Issues
- Resolves #352 

## Testing
- Unit tests have been updated to account for the change to the base PGE schema.
- The CSLC SAS image now utilizes Python 3.11, which changes the default behavior for how `IntEnum` values (such as our `ErrorCode` enum) are represented as strings. An affected test in `test_logger.py` has been updated to support both Python 3.11 and previous.

### DIFF
--- a/.ci/scripts/build_cslc_s1.sh
+++ b/.ci/scripts/build_cslc_s1.sh
@@ -24,7 +24,7 @@ BUILD_DATE_TIME=$(date -u +'%Y-%m-%dT%H:%M:%SZ')
 # defaults, SAS image should be updated as necessary for new image releases from ADT
 [ -z "${WORKSPACE}" ] && WORKSPACE=$(realpath $(dirname $(realpath $0))/../..)
 [ -z "${TAG}" ] && TAG="${USER}-dev"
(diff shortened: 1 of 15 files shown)


JimHofman (approved):
Looks great.  Pretty straight forward.
Is the long range development/maintenance plan to stay current with python releases?  There seem to be some significant speed enhancements for 3.11.  Maybe something for us to look at.
[](https://www.python.org/downloads/release/python-3110/)

david-inglish (approved):
I don't see any issues with the changes.

Issue #348, opened 2023-08-21
DISP-S1 ISO XML Creation


collinss-jpl:
Ticket to add capability to produce the ISO XML metadata file by instantiating the jinja2 template with extracted product metadata.

For DISP-S1, we should currently only expect a single ISO xml file to be instantiated, since we only expect one output product for the job. This means that the paradigm will be closer to the DSWx-HLS PGE, where all that is needed is an implementation for `_create_iso_metadata()` within the `DispS1PostProcessorMixin` class. The function should be responsible for both gathering the constituent metadata dictionaries the template expects, and then returning the contents of the rendered template.

collinss-jpl:
Closed by #367

PR #367, opened 2023-09-13
#348 add disp_s1 iso-metadata

JimHofman:
## Description
- Produce ISO XML metadata using jinja1 template

## Testing
- added a unit test (98% code coverage)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,6 +5,7 @@ h5py
 Jinja2
 jsonschema
 mgrs
+numpy
 PyYAML
 pylint
 pytest
--- a/src/opera/pge/disp_s1/disp_s1_pge.py
+++ b/src/opera/pge/disp_s1/disp_s1_pge.py
@@ -9,6 +9,7 @@ from Sentinel-1 A/B (S1-A/B) data.
 """
 
 import glob
+import os.path
 import re
 from collections import OrderedDict
 from os import listdir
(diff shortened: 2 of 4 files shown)


src/opera/pge/disp_s1/disp_s1_pge.py
```diff
@@ -139,6 +142,106 @@ def _validate_output(self):
...
+        use in filling out the ISO metadata template for the DISP-S1 PGE.
+
+        Returns
+        -------
+        output_product_metadata : dict
+            Dictionary containing DISP-S1 output product metadata, formatted
+            for use with the ISO metadata Jinja2 template.
+
+        """
+        # Find a single representative output DSWx-HLS product, they should all
```
david-inglish, line 157: DSWx-HLS -> DISP-S1

src/opera/pge/disp_s1/disp_s1_pge.py
```diff
@@ -139,6 +142,106 @@ def _validate_output(self):
...
+            Dictionary containing the custom metadata as expected by the ISO
+            metadata Jinja2 template.
+
+        """
+        custom_metadata = {
+            'ISO_OPERA_FilePackageName': self._core_filename(),
+            'ISO_OPERA_ProducerGranuleId': self._core_filename(),
+            'MetadataProviderAction': "creation",
+            'GranuleFilename': self._core_filename(),
+            'ISO_OPERA_ProjectKeywords': ['OPERA', 'JPL', 'DSWx', 'Dynamic', 'Surface', 'Water', 'Extent'],
```
david-inglish, line 196: Should these particular keywords be related to surface displacement?

requirements.txt
collinss-jpl, line 1: Was this intended for inclusion? I want to keep the versions un-pinned in this file since we install into an existing environment within the SAS container.

collinss-jpl (changes requested):
Implementation looks good, but the unit test needs a few more checks, and the changes to requirements.txt need to be reverted before merge.

collinss-jpl (approved):

PR #369, opened 2023-09-18
RTC-S1 SAS Final PGE Integration

### LINKED ISSUE #362: RTC-S1 Static Layer Data Access URL Metadata Injection
As part of a recent initiative, its been decided that the RTC-S1 output products will conform to "[CEOS metadata compliance](https://ceos.org/ourwork/workinggroups/wgiss/access/opensearch/)". What this means practically is that the RTC-S1 product metadata will contain several additional fields to provide linkage to things like the source data access, product data access and  static layer product access.

This ticket is only concerned with formation of the "static layer data access" URL, which must be injected into the RTC-S1 product metadata after the SAS has completed processing. This is because the URL contains a reference to specific burst ID's, which the PGE is best suited to work with.

The format template of the URL is as follows:

`https://search.asf.alaska.edu/#/?dataset=OPERA-S1&productTypes=RTC-STATIC&burstID=<burst_id>&productVersion=<product ver>`

Where `<burst_id>` and `<product_ver>` will be filled in by the PGE.

The goal of this ticket is to add a function to the RTC-S1 PGE Post-processor to iterate over all `.h5` metadata products, extract `<burst id>` and `<product ver>`, and use them to instantiate the URL template, then write the instantiated URL into the product metadata and save it back out to disk. This can occur after products have been assigned file name conventions.

The hdf5 metadata path to write the URL to is **TBD**

### LINKED ISSUE #364: RTC-S1 SAS Final PGE Integration
Ticket to track integration of the RTC-S1 Final SAS delivery with the PGE library.

In addition to the basic reintegration steps, this ticket should also include any updates needed to the ISO XML template to make sure all appropriate metadata is being collected, and that the integration test pipeline is updated with the latest assets for the Final delivery.

### LINKED ISSUE #365: RTC-S1 Static Layer Bounding Box Extraction
Currently, we use the HDF5 metadata output product to populate the ISO XML template for both the standard and static layer workflows for RTC-S1. However, for the static layer workflow the RTC-S1 SAS does not provide a bounding box polygon that describes the spatial extent of the static layer (which differs from the extent of the standard burst product). This results in the bounding polygon being missing from the ISO XML for static layer products.

As a workaround, ADT has informed us that we can extract a similar bounding box from one of the GeoTIFF static layer products directly, rather than from the .h5 product. The purpose of this ticket is to update the metadata extraction code for RTC-S1 to extract the bounding box from a GeoTIFF product when running the static layer workflow (which can be determined by the RunConfig). The existing GeoTIFF metadata extraction function should be usable. The bounding box should be found under the `BOUNDING_BOX` key.

collinss-jpl:
## Description
- This branch integrates the Final SAS delivery for RTC-S1 with the PGE library. Expected assets for the integration test have been updated, as well as the example RunConfig. This branch also adds updates to support usage of the product bounding box in the ISO XML metadata for static layer products, and adds an additional post-processing step to ensure that the static layer data access URL is injected properly into the output metadata products for the baseline RTC-S1 workflow.

## Affected Issues
- Resolves #362 
- Resolves #364 
- Resolves #365 

## Testing
- Unit tests have been updated to add tests specific to the static layer workflow. Tests have also been added for the URL injection step.

### DIFF
--- a/examples/rtc_s1_sample_runconfig-v2.0.0.yaml
+++ b/examples/rtc_s1_sample_runconfig-v2.0.0.yaml
@@ -1,4 +1,4 @@
-# Sample RunConfig for use with the RTC-S1 PGE v2.0.0-rc.2.1
+# Sample RunConfig for use with the RTC-S1 PGE v2.0.0
 # This RunConfig should require minimal changes in order to be used with the
 # OPERA PCM.
 
@@ -35,7 +35,7 @@ RunConfig:
                     dem_file: /home/rtc_user/input_dir/dem.tif
 
                     # Burst database, must be an .sqlite3 file
-                    burst_database_file: /home/rtc_user/input_dir/burst_db_20230713-bbox-only.sqlite
+                    burst_database_file: /home/rtc_user/input_dir/burst_db_0.2.0_230831-bbox-only.sqlite
 
             ProductPathGroup:
                 # Path to where output products should be stored
@@ -58,7 +58,7 @@ RunConfig:
                 ProductIdentifier: RTC_S1
 
                 # Product version specific to output products
-                ProductVersion: 1.0
+                ProductVersion: "1.0"
 
                 # Path to the executable to run, path must be reachable from
                 # within the Docker container (i.e. findable with a "which" command)
@@ -96,7 +96,7 @@ RunConfig:
                 # RTC static layer product(s) should be considered valid.
                 # This field must be provided for RTC-S1 jobs when static layer
                 # generation is enabled (see below), and must be of the form YYYYMMDD
-                DataValidityStartDate: 20140101
+                DataValidityStartDate: 20140403
 
             QAExecutable:
                 # Set to True to enable execution of an additional "Quality Assurance"
@@ -140,19 +140,21 @@ RunConfig:
                             - /home/rtc_user/input_dir/S1B_OPER_AUX_POEORB_OPOD_20180524T110543_V20180503T225942_20180505T005942.EOF
                         # Burst IDs to process. Leave empty to process all bursts.
                         burst_id:
+                        # Location from where the source data can be retrieved (URL or DOI)
+                        source_data_access: "https://search.asf.alaska.edu/#/?dataset=SENTINEL-1&productTypes=SLC"
 
                     # This section should match the DynamicAncillaryFilesGroup of the PGE RunConfig
                     dynamic_ancillary_file_group:
                         # Digital elevation model
                         dem_file: /home/rtc_user/input_dir/dem.tif
-                        dem_file_description: Digital Elevation Model (DEM) for the NASA OPERA project (v1.0) based on the Copernicus DEM 30-m and Copernicus 90-m referenced to the WGS84 ellipsoid
+                        dem_file_description: "Digital Elevation Model (DEM) for the NASA OPERA project version 1.1 (v1.1) based on the Copernicus DEM 30-m and Copernicus 90-m referenced to the WGS84 ellipsoid"
 
                     static_ancillary_file_group:
                         # burst database sqlite file
-                        burst_database_file: /home/rtc_user/input_dir/burst_db_20230713-bbox-only.sqlite
+                        burst_database_file: /home/rtc_user/input_dir/burst_db_0.2.0_230831-bbox-only.sqlite
 
                     product_group:
-                        product_version: 1.0
+                        product_version: "1.0"
 
                         # This should match the path used for OutputProductPath
                         product_path: /home/rtc_user/output_dir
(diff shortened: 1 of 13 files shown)


examples/rtc_s1_sample_runconfig-v2.0.0.yaml
```diff
@@ -163,16 +165,14 @@ RunConfig:
...
-                        product_id: OPERA_L2_RTC-S1_{burst_id}
-
                         # Validity start date for RTC-S1-STATIC products in the format YYYYMMDD,
                         # This field must be provided for RTC-S1 jobs when static layer
                         # generation is enabled, and should match the value assigned to
                         # DataValidityStartDate above.
-                        rtc_s1_static_validity_start_date: 20140101
+                        rtc_s1_static_validity_start_date: 20140403
+
+                        # Location from where the output product can be retrieved (URL or DOI)
```
JimHofman, line 174: Just curious.  This value matches 'source_data_access' under 'input_file_group'.  If this will always be the case (which I imagine it will not be, but just throwing it out there), it might make sense to factor it out into another key.

src/opera/pge/rtc_s1/templates/OPERA_ISO_metadata_L2_RTC_S1_template.xml.jinja2
```diff
@@ -1238,18 +1248,18 @@
...
                                             </eos:description>
                                             <eos:dataType>
                                                 <eos:EOS_AdditionalAttributeDataTypeCode codeList="https://cdn.earthdata.nasa.gov/iso/resources/Codelist/eosCodelists.xml#EOS_AdditionalAttributeDataTypeCode" codeListValue="boolean">boolean</eos:EOS_AdditionalAttributeDataTypeCode>
                                             </eos:dataType>
                                         </eos:EOS_AdditionalAttributeDescription>
                                     </eos:reference>
                                     <eos:value>
-                                        <gco:CharacterString>{{ product_output.processingInformation.parameters.dryTroposphericGeolocationCorrectionApplied }}{# ISO_OPERA_dryTroposphericGeolocationCorrectionApplied #}</gco:CharacterString>
+                                        <gco:CharacterString>{{ product_output.processingInformation.parameters.staticTroposphericGeolocationCorrectionApplied }}{# ISO_OPERA_dryTroposphericGeolocationCorrectionApplied #}</gco:CharacterString>
```
david-inglish, line 1262: {# ISO_OPERA_dryTroposphericGeolocationCorrectionApplied #} -> {# ISO_OPERA_staticTroposphericGeolocationCorrectionApplied #}

src/opera/pge/rtc_s1/rtc_s1_pge.py
```diff
@@ -710,6 +716,40 @@ def _collect_rtc_product_metadata(self, metadata_product):
         output_product_metadata['data']['length'] = len(output_product_metadata['data']
                                                         ['yCoordinates'])
 
+        product_type = self.runconfig.product_type
+
+        # If creating ISO metadata for a static layer product, we need to
```
JimHofman, line 721: I think this comment needs to be expanded, particularly since we now have 2 ways of extracting data to determine the bounding boxes.

src/opera/pge/rtc_s1/rtc_s1_pge.py
```diff
@@ -710,6 +716,40 @@ def _collect_rtc_product_metadata(self, metadata_product):
...
+            lat_min, lat_max, lon_min, lon_max = translate_utm_bbox_to_lat_lon(bounding_box, epsg_code)
+
+            output_product_metadata['staticBoundingBox'] = {
+                'lonMin': lon_min,
+                'latMin': lat_min,
+                'lonMax': lon_max,
+                'latMax': lat_max
+            }
+
+        # We also need to assign the URL for static layer data access for
```
JimHofman, line 736: Again, the comment (smiling).  This statement ending makes me thing 'as opposed to ......'

david-inglish (approved):
I just see the one minor thing I commented on.

src/opera/test/data/test_rtc_s1_static_config.yaml
```diff
@@ -0,0 +1,104 @@
...
+                    primary_executable:
+                        product_type: RTC_S1_STATIC
+
+                    product_group:
+                        processing_type: NOMINAL
+                        product_version: 1.0
+                        product_path: rtc_s1_test/output_dir
+                        scratch_path: rtc_s1_test/scratch_dir
+                        output_dir: rtc_s1_test/output_dir
+                        product_id: rtc_product
```
JimHofman, line 82: Similarly, do you need 'product_data_access: <some value>' here?

src/opera/test/data/test_rtc_s1_static_config.yaml
```diff
@@ -0,0 +1,104 @@
...
+                        pge_name: RTC_S1_PGE
+
+                    input_file_group:
+                        safe_file_path:
+                            - rtc_s1_test/input_dir/SAFE.zip
+                        orbit_file_path:
+                            - rtc_s1_test/input_dir/ORBIT.EOF
+                        burst_id:
+                            - t069_147170_iw1
+
```
JimHofman, line 66: Do you need  'source_data_access' after 'burst_id', or is it something that we don't need to test?

JimHofman (commented):
A couple of comments
Nice job Scott, that was a lot of work.

Issue #328, opened 2023-07-31
pylint/flake8 cleanup for R3


collinss-jpl:
Been awhile since we've burned down the pylint/flake8 for the current R3 repo, so this sprint would be a good opportunity to resolve or suppress as many findings as feasible.

Recent findings for each PGE container build can be found in the Jenkins build/test pipeline for [315_isoMetaData](https://opera-pge-ci.jpl.nasa.gov/job/opera-pge-build-test-pipeline/job/315_isoMetaData/).

Of particular note are the findings for metadata_util.py for both flake8 and pylint. Since we have a lot of test code to create dummy metadata products, we need to add the appropriate statments or pragmas to disable static analysis on the various `create_test_<pge>_metadata_product`. This will go a long way in trimming the list of findings, so only the legitimate ones should remain.

collinss-jpl:
Closed by #373

PR #373, opened 2023-10-05
#328 - flake8/pylint for review

JimHofman:
## Description
- Cleaned things up a bit.  All the flake8 commands are clean, and I adding a few pylint commands to clean things up too.
- We are at 95% total coverage with only img_utils.py  just below 90% at 89%, I left that alone (other than generaral cleanup) since it's about to be factored into another module.  Make 'plyint' changes for  in .pylintrc and in metadata_utils.py and input_validation.py.

## Testing
- All unit tests run successfully

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -14,6 +14,10 @@ ignore =
     D404
     # E722: do not use bare 'except'
     E722
+    # W503: line break before binary operator
+    W503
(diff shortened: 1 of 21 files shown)


collinss-jpl (approved):
Changes look great! I just nudged some of the disable directives around so they're more locally scoped to problematic functions/classes.

Issue #359, opened 2023-09-06
DSWx-S1 Beta SAS Acceptance Test


sdcolli1:
Ticket to track acceptance testing for the DSWx-S1 SAS Beta delivery.

The container tarball for the delivery can be found in Artifactory-FN under the path `general/gov/nasa/jpl/opera/adt/dswx_s1/r2.1/beta/dockerimg_dswx_s1_beta_0.2.1.tar`, test inputs and expected outputs can be found under `general/gov/nasa/jpl/opera/adt/dswx_s1/r2.1/beta/sample_data.zip`. Documentation for the delivery is under `general/gov/nasa/jpl/opera/adt/dswx_s1/r2.1/beta/documents/`

A new wiki page needs to be created to capture the Acceptance Testing procedure for the Beta delivery. The procedure for the previous Interface delivery may be used as a starting point. However, unlike the previous Interface delivery, a script to compare the results of the Acceptance Test with the expected output products (within `sample_data.zip`) is now available, and its use needs to be incorporated into the testing procedure.

A copy of the DSWx-S1 comparison script can be obtained here: https://raw.githubusercontent.com/opera-adt/DSWX-SAR/main/src/dswx_sar/dswx_comparison.py

It is up to the developer to decide whether the `dswx_comparison.py` script should be executed from within the container (mounting the Acceptance Test outputs and expected outputs into the container), or from outside the container as a standalone Python process.

collinss-jpl:
Closed by #387

Issue #311, opened 2023-06-27
Integrate comparison script usage for DSWx-S1 integration test


collinss-jpl:
At the time of development of the integration test pipeline for DSWx-S1, no comparison script had been provided by ADT to compare output products to the golden expected. This ticket is to integrate such a script, once it has been furnished by ADT.

Update: the comparison script has been delivered with the DSWx-S1 Beta delivery. A copy can be obtained here: https://raw.githubusercontent.com/opera-adt/DSWX-SAR/main/src/dswx_sar/dswx_comparison.py

~The task for this ticket is to develop some shell script code invokes the dswx_comparsion.py script, either via `docker run` to invoke within the dswx-s1 container, or by committing a copy of dswx_comparison.py to the PGE repo and calling it locally.~

Once dswx_compare.py is available in the repository, it needs to be integrated into test_int_dswx_s1.sh here: https://github.com/nasa/opera-sds-pge/blob/main/.ci/scripts/test_int_dswx_s1.sh#L162-L163

Finally, once the call to dswx_compare.py is added to test_int_dswx_s1.sh, the following lines should also be un-commented to complete the stats collection HTML file: https://github.com/nasa/opera-sds-pge/blob/main/.ci/scripts/test_int_dswx_s1.sh#L171-L172

collinss-jpl:
Closed in #390

PR #387, opened 2023-11-28
359 dswx s1 comparison script

rjbdev05:
## Description
- <Provide a description of the changes/updates/fixes provided by the branch>

## Affected Issues
- <#359 - DSWx-S1-Beta SAS Acceptance Test>
- <#311 - Integrate comparison script usage for DSWx-S1 Integration Test>

## Testing
- <Remove the background metrics collection>
- <Once the PR is created, don't forget to check the Build/Test Jenkins pipeline for your branch to ensure all tests are passing before merge>

### DIFF
--- a/.ci/scripts/diff_dswx_files.py
+++ b/.ci/scripts/diff_dswx_files.py
@@ -1,28 +1,19 @@
-#   diff_dswx_files.py
-#
-#   For: OPERA project
-#
-#   11-03-2023 - Ray Bambery
-#       initial release
-#
-#   Description: Python program to perform Quality Assurance Test
-#       on a directory of NETcdf4 files - searches for files with
-#       a .tif file extension.
-#       This script calls a subprocess script,  dswx_comparison.py
-#       located in the dswx_s1_x.y.z directory shown as <DSWX-S1_DIR>
-#       in the wiki page 
-#       https://github.com/nasa/opera-sds-pge/wiki/DWSX_S1_Beta-Acceptance-Test-Instruuctions
-#       which ensures there are matching files in each input directory
-##
-#   Arguments:
-#       expected_output_dir output_dir dir
-#
-#   Outputs: A terminal stream of  results of the tests showing 
-#       PASS (in GREEN) or FAIL (in RED)
-#   
-#   Calls:  
-#       dswx_comparison.py
-#
+#!/usr/bin/env python3
+""" 
+==================
+diff_dswx_files.py
+==================
+
+   Description: Python program to perform Quality Assurance Test
+       on a directory of NETcdf4 files - searches for files with
+       a .tif file extension.
+       This script calls a subprocess script,  dswx_comparison.py
+       located in the dswx_s1_x.y.z directory shown as <DSWX-S1_DIR>
+       in the wiki page 
+       https://github.com/nasa/opera-sds-pge/wiki/DWSX_S1_Beta-Acceptance-Test-Instruuctions
+       which ensures there are matching files in each input directory.
+"""
+
 import os
 import sys
 import subprocess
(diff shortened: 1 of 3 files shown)


.ci/scripts/diff_dswx_files.py
```diff
@@ -0,0 +1,139 @@
...
+            sys.exit(-1)
+        else:
+            print ("sys.argv[1] = ",sys.argv[1])
+            print ("sys.argv[2] = ",sys.argv[2])
+#           options = parser.parse_args()
+#           print ("options are ", options)
+
+    return sys.argv
+#################################################################
+def get_files(options):
```
collinss-jpl, line 60: This is as good a place as any to describe the standards we use for function docstrings:

As much as possible, I want to see docstrings that follow the "numpy" convention for documentation. The general template for a numpy-style docstring looks like the following:

```python
def function_name(arg1, arg2)
    """
    <The first sentence/paragraph of the docstring should explain the purpose of the function at a high-level.>
 
    Notes
    ------
    <Any additional documentation worth calling out that is not provided by the main descriptive sentence/paragraph>

    Parameters
    ------------
    arg1 : <argument type>
       <Description of the argument, including any constraints on value or invariants upheld by the function>
    arg2: <argument type>
       <Similar description for arg2>

    Raises
    -------
    <Exception type>
         <If the function can raise an exception, document the circumstances that would cause it here>

    Returns
    --------
    result : <type of returned value or values>
         <Description of the returned value or values>
    """
  
```
Note that if the function does not take any parameters, a `Parameters` section is not necessary. Similarly, if the function does not return anything, no `Returns` section is necessary, and if no exceptions are **directly** raised by the function, no `Raises` section is necessary.

For very simple functions, the docstring can be confined to a single line beneath the `def`:

```python
def simple_function():
    """Example single line docstring describing the simple function"""
```

You can find some good examples of our documentation practices within the following modules:
https://github.com/nasa/opera-sds-pge/blob/main/src/opera/util/run_utils.py
https://github.com/nasa/opera-sds-pge/blob/main/src/opera/util/tiff_utils.py

For the sake of practice, I would like to see docstings like the ones I've described added to the `_parse_args` and `get_files` functions within this module. Don't worry about `dswx_comparison.py` since it already has similar(ish) docstrings already.

.ci/scripts/diff_dswx_files.py
```diff
@@ -0,0 +1,139 @@
...
+#       dswx_comparison.py
+#
+import os
+import sys
+import subprocess
+import argparse
+import glob
+#import netCDF4 as nc
+
+##################################################################
```
collinss-jpl, line 33: I would prefer these kinds of dividers be removed, two lines of whitespace is sufficient for separating functions:

```
def func1():
    pass


def func2():
    pass
```

.ci/scripts/diff_dswx_files.py
```diff
@@ -0,0 +1,139 @@
...
+    expected_dir = options[1]
+    output_dir = options[2]
+
+    # list to store files
+    exp = [os.path.basename(f) for f in sorted(glob.glob(os.path.join(expected_dir, '*.tif')))]
+#    print (exp)
+    expected_file_count = len(exp)
+    print("expected_file_count = ",len(exp))
+
+    ecnt = 0
```
collinss-jpl, line 75: I would like to see longer, more descriptive variable names used wherever possible so its easier for other developers to maintain. So in this case, I would rename `enct` to `expected_counter`, `ocnt` to `output_counter` and so on.

.ci/scripts/diff_dswx_files.py
```diff
@@ -0,0 +1,139 @@
...
+        print ("[FAIL]  expected file_count == 0")
+        print ("    Expected file count of 0 usually implies a typo in directory name")
+        sys.exit(-1)
+    if output_file_count == 0:
+        print ("[FAIL]  output file_count == 0")
+        print ("    Output file count of 0 usually implies a typo in directory name")
+        sys.exit(-1)
+
+    if output_file_count > expected_file_count:
+        print ("[FAIL]  output_file_count ",output_file_count," exceeds expected_file_count ",expected_file_count)
```
collinss-jpl, line 112: Please include a single whitespace after every comma

.ci/scripts/diff_dswx_files.py
```diff
@@ -0,0 +1,139 @@
...
+        sys.exit(-1)
+    if expected_file_count > output_file_count:
+        print ("[FAIL]  expected_file_count ",expected_file_count," exceeds output_file_count ",output_file_count)
+        sys.exit(-1)
+
+#$ could be ocnt
+    for i in range(0, ecnt):
+#        print (i)
+        print ("python3 dswx_comparison.py  {}  {}".format(expected_dir + '/' + exp[i],output_dir + '/' + out[i]))
+        os.system("python3 dswx_comparison.py  {}  {}".format(expected_dir + '/' + exp[i],output_dir + '/' + out[i]))
```
collinss-jpl, line 122: Couple of notes here:

1. When assembling a file path, it is better to use a function like [os.path.join](https://docs.python.org/3/library/os.path.html#os.path.join) or the [Pathlib](https://docs.python.org/3/library/pathlib.html#basic-use) module, rather than string concatenation.
2. Use of `str.format()` has been deprecated as of late, and the current best practice is to utilize[ F-strings](https://realpython.com/python-f-strings/#doing-string-interpolation-with-f-strings-in-python) since they make for more readable code.

For the sake of a good practice activity, please rewrite lines 121 and 122 to utilize the conventions I outlined above.

.ci/scripts/diff_dswx_files.py
```diff
@@ -0,0 +1,139 @@
+#   diff_dswx_files.py
```
collinss-jpl, line 1: Since this is expected to be a callable script, the first line should be a [shebang](https://en.wikipedia.org/wiki/Shebang_(Unix)) that instructs the OS on how the script should be executed.

For Python, this is `#!/usr/bin/env python3`

.ci/scripts/diff_dswx_files.py
```diff
@@ -0,0 +1,139 @@
...
+#
+#   For: OPERA project
+#
+#   11-03-2023 - Ray Bambery
+#       initial release
+#
+#   Description: Python program to perform Quality Assurance Test
```
collinss-jpl, line 8: I like this high-level description of the script, but we typically format such descriptions the following way:

```
"""
==================
diff_dswx_files.py
==================

Python program to perform Quality Assurance Test on a directory of NETcdf4 files - searches for files with a .tif file extension.

This script calls a subprocess script, dswx_comparison.py located in the dswx_s1_x.y.z directory shown as <DSWX-S1_DIR> in the wiki page 
https://github.com/nasa/opera-sds-pge/wiki/DWSX_S1_Beta-Acceptance-Test-Instruuctions which ensures there are matching files in each input directory

"""

The other information, like the `Arugments`, `Outputs`, and `Calls`, are typically allocated to the docstrings for functions themselves (see my comment about numpy-style docstrings).

```

.ci/scripts/diff_dswx_files.py
```diff
@@ -0,0 +1,139 @@
+#   diff_dswx_files.py
+#
+#   For: OPERA project
+#
+#   11-03-2023 - Ray Bambery
+#       initial release
```
collinss-jpl, line 6: Change history such as this is captured within the commit history in git itself, so you do not need to document it within the code itself.

collinss-jpl (changes requested):
Overall the code looks good! It does exactly what it needs to do. But since this is your first pull request, and first contribution of code to the OPERA PGE repo, I focused my review on coding and documentation standards that will ensure your code has the same "look and feel" as the rest of our repository.

Please address the following comments in your working branch and push the changes to `origin` when they are ready. This pull request will update itself automatically to any changes that are pushed. If you have any questions about anything, you can reply directly to each of the comments to start a thread, and I will respond as soon as I can. Thanks @rjbdev05!

.ci/scripts/diff_dswx_files.py
```diff
@@ -0,0 +1,141 @@
+#!/usr/bin/env python3
+""" 
+==================
+diff_dswx_files.py
+==================
+
+   Description: Python program to perform Quality Assurance Test
```
JimHofman, line 7: We are using Docstring for all our python code.

collinss-jpl (approved):
Thank you for addressing my feedback @rjbdev05. I noticed a few other minor changes needed, but for the sake of getting this branch merged I went ahead an pushed the changes myself. You can look at commit e6626c8338334fa73fbdb1071911114a15d0d6a6 to see the changes I made.

PR #389, opened 2024-01-03
381 grib to netcdf

JimHofman:
## Description
- Added translation from .grb (grib) files to .nc (netcdf) to disp_s1_pge

## Affected Issues
- None

## Testing
- Added unit test to test temporary sas config file created during run time.
- flake8, pytest

### DIFF
--- a/src/opera/pge/disp_s1/disp_s1_pge.py
+++ b/src/opera/pge/disp_s1/disp_s1_pge.py
@@ -68,7 +68,7 @@ class DispS1PreProcessorMixin(PreProcessorMixin):
 
     def convert_troposphere_model_files(self):
         """
-        Convert grip (.grp) files to netCDF (.nc)
+        Convert grib (.grb) files to netCDF (.nc)
         Update the in-memory runconfig object such that the SAS/dynamic_ancillary_file_group/troposphere_files
         section now points to the converted .nc files in the scratch directory.
 
@@ -82,9 +82,9 @@ class DispS1PreProcessorMixin(PreProcessorMixin):
         netcdf_file_list = []
 
         for tropo_file in troposphere_model_files_list:
-            if tropo_file[-3:] == 'grb':
+            if splitext(tropo_file)[-1] == '.grb':
                 # change the extension to .nc
-                netcdf_file = join(scratch_dir, tropo_file.split('/')[-1][:-4] + '.nc')
+                netcdf_file = join(scratch_dir, splitext(basename(tropo_file))[0] + '.nc')
                 # This list of will the new paths to the converted files in the in-memory runconfig file.
                 netcdf_file_list.append(netcdf_file)
                 grib_file_name = tropo_file
--- a/src/opera/pge/disp_s1/disp_s1_pge.py
+++ b/src/opera/pge/disp_s1/disp_s1_pge.py
@@ -12,6 +12,7 @@ from Sentinel-1 A/B (S1-A/B) data.
 
 import glob
 import os.path
+import subprocess
 from collections import OrderedDict
 from os import listdir
 from os.path import abspath, basename, exists, getsize, join, splitext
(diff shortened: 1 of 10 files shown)


src/opera/pge/disp_s1/disp_s1_pge.py
```diff
@@ -63,6 +64,45 @@ def run_preprocessor(self, **kwargs):
                                              self.runconfig.algorithm_parameters_schema_path,
                                              self.runconfig.algorithm_parameters_file_config_path,
                                              self.logger)
+        self.convert_troposphere_model_files()
+
+    def convert_troposphere_model_files(self):
+        """
+        Convert grip (.grp) files to netCDF (.nc)
```
collinss-jpl, line 71: grip should be "grib" and .grp should be .grb

src/opera/pge/disp_s1/disp_s1_pge.py
```diff
@@ -63,6 +64,45 @@ def run_preprocessor(self, **kwargs):
...
+        # Retrieve the troposphere weather model file group from the run config file
+        troposphere_model_files_list = self.runconfig.sas_config['dynamic_ancillary_file_group']['troposphere_files']
+
+        # Converted files will be stored in the scratch directory.
+        scratch_dir = self.runconfig.sas_config['product_path_group']['scratch_path']
+
+        netcdf_file_list = []
+
+        for tropo_file in troposphere_model_files_list:
+            if tropo_file[-3:] == 'grb':
```
collinss-jpl, line 85: I would prefer if this leveraged `os.path.splitext`: `if os.path.splitext(tropo_file)[-1] == '.grb'`

src/opera/pge/disp_s1/disp_s1_pge.py
```diff
@@ -63,6 +64,45 @@ def run_preprocessor(self, **kwargs):
...
+                # change the extension to .nc
+                netcdf_file = join(scratch_dir, tropo_file.split('/')[-1][:-4] + '.nc')
```
collinss-jpl, line 87: `splitext` could be similarly leveraged here: `netcdf_file = join(scratch_dir, os.path.splitext(tropo_file)[0] + '.nc')`

src/opera/test/data/test_disp_s1_config.yaml
```diff
@@ -25,6 +25,7 @@ RunConfig:
                         - disp_s1_pge_test/input_dir/jplg0410.18i.Z
                     troposphere_files:
                         - disp_s1_pge_test/input_dir/GMAO_tropo_20180210T000000_ztd.nc
+                        - disp_s1_pge_test/input_dir/GMAO_tropo_20180211T000000_ztd.nc
```
collinss-jpl, line 28: For consistency, these should be set to the same set of filenames used in the SAS/dynamic_ancillary_file_group/troposphere_files section below

src/opera/test/pge/disp_s1/test_disp_s1_pge.py
```diff
@@ -28,6 +31,30 @@
 from opera.util.input_validation import validate_disp_inputs
 
 
+def mock_grib_to_netcdf(*popenargs, input=None, capture_output=False, timeout=None, check=False, **kwargs):
+    """
+    Mack grit_to_netcdf function.
```
collinss-jpl, line 36: Should read "Mock grib_to_netcdf function", but what this function really is is a mock implementation for `subprocess.run`, and that should be noted in the documentation here.

src/opera/test/pge/disp_s1/test_disp_s1_pge.py
```diff
@@ -823,6 +857,45 @@ def test_disp_s1_pge_validate_product_output(self):
...
+        pge.run_preprocessor()
+        pge.run_sas_executable()
+        temp_sas_runconfig = 'disp_s1_pge_test/scratch_dir/' + pge.runconfig.filename.split('/')[-1][:-5] + '_sas.yaml'
+        # open the yaml file
+        with open(temp_sas_runconfig, 'r') as file:
+            data = yaml.safe_load(file)
+        ending_tropo_paths = data['dynamic_ancillary_file_group']['troposphere_files']
+        # Verify the .grb files are in /scratch_dir
+        # Strip the path and extension (changed to .nc in this case) to allow comparison of file name only.
+        for i in ending_tropo_paths:
```
collinss-jpl, line 892: I think we could add one more tested condition here, namely that each path in `ending_tropo_paths` also exists on disk, which would basically be testing that the mock implementation is working as expected. Something like: `self.assertTrue(os.path.exists(ending_tropo_path))`

src/opera/test/pge/disp_s1/test_disp_s1_pge.py
```diff
@@ -823,6 +857,45 @@ def test_disp_s1_pge_validate_product_output(self):
...
+
+        pge = DispS1Executor(pge_name="DispS1PgeTest", runconfig_path=runconfig_path)
+        # Pull out the starting value of the troposphere_files
+        with open(runconfig_path, 'r', encoding='utf-8') as infile:
+            runconfig_dict = yaml.safe_load(infile)
+        starting_tropo_paths = \
+            runconfig_dict['RunConfig']['Groups']['SAS']['dynamic_ancillary_file_group']['troposphere_files']
+        # Strip the path and extension (.grb) in order to compare just the file names after
+        # conversion and placement of the converted file into /scratch_dir/<fname>.nc
+        for i in starting_tropo_paths:
```
collinss-jpl, line 879: Even though this is just test code, I would still like to avoid single-letter loop variables, so could you please rename this to `for starting_tropo_path in starting_tropo_paths`, or similar.

src/opera/test/pge/disp_s1/test_disp_s1_pge.py
collinss-jpl, line 892: Same comment as above regarding single letter loop variables

collinss-jpl (changes requested):
Implementation looks good, just had a few mostly stylistic comments I would like to see addressed before this is merged.

src/opera/pge/disp_s1/disp_s1_pge.py
JimHofman: Yeah, more flexible in most cases.

collinss-jpl (approved):

PR #390, opened 2024-01-08
comparison test for ouput and comparison files

rjbdev05:
## Description
- Modified test script .ci/script/test_int_dswx_s1.sh by adding dswx_comparison.py into indicated line

## Affected Issues
- #311

## Testing
- ran test test_int_dswx_s1.sh with output to /data/rjb/tmp/tmp.XCeU5RLzCM

### DIFF
--- a/.ci/scripts/test_int_dswx_s1.sh
+++ b/.ci/scripts/test_int_dswx_s1.sh
@@ -115,21 +115,16 @@ else
     # Compare output files against expected files
     for output_file in "$output_dir"/*
     do
-        docker_out="N/A"
+        compare_output="N/A"
         compare_result="N/A"
         expected_file="N/A"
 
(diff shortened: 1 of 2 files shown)


.ci/scripts/test_int_dswx_s1.sh
```diff
@@ -43,7 +43,7 @@ test_int_setup_data_tmp_directory
 test_int_setup_test_data
 
 # Setup cleanup on exit
-trap test_int_trap_cleanup EXIT
+#trap test_int_trap_cleanup EXIT
```
collinss-jpl, line 46: This needs to be re-enabled (remove leading #) before this branch is merged

collinss-jpl (approved):
Looks good, I just had one comment about re-enabling the cleanup trap function before this can get merged

PR #397, opened 2024-01-23
386 ci dir cleanup

JimHofman:
## Description
- Refactor directories in .ci, group by pge or function.  Modify Jenkinsfiles for the directory changes.   Miscellaneous code clean up

## Affected Issues
- Will probably be some manual merging with main.

## Testing
- All the modified '_test.sh' scripts ran correctly
- Jenkins runs were satisfactory.


.ci/scripts/dswx_hls/build_dswx_hls.sh
```diff
@@ -32,7 +32,7 @@ echo "TAG: $TAG"
 echo "SAS_IMAGE: $SAS_IMAGE"
 
 # Check that the .ci scripts directory exists
-if [ ! -d "${WORKSPACE}/.ci" ]; then
+if [ ! -d "${WORKSPACE}" ]; then
```
collinss-jpl, line 35: The other build scripts have this same check on `${WORKSPACE}/.ci`, so why was it only changed for DSWx-HLS?

.ci/scripts/py_util/diff_dswx_files.py
collinss-jpl, line 1: Since this script is actually specific to DSWx-S1, I would like to see it co-located with the other DSWx-S1 scripts under `.ci/scripts/dswx_s1`. Once its moved we can assign it a more descriptive name in a future ticket. Note that `test_int_dswx_s1.sh` will likely need to be updated as well to point to this new location.

.ci/scripts/py_util/disp_validate_product_opera_pge.py
collinss-jpl, line 1: Similar to comment on `diff_dswx_files.py`, this script should be moved to `.ci/scripts/disp_s1`

.ci/scripts/py_util/dswx_compare_opera_pge.py
collinss-jpl, line 1: Let's move this to `.ci/scripts/dswx_hls`. Note that `test_int_dswx_hls.sh` will likely need to be updated as well to point to this new location.

.ci/scripts/py_util/dswx_comparison.py
collinss-jpl, line 1: This script should also be moved to `.ci/scripts/dswx_s1`. I think once you've moved this and the others noted, we will not need the `py_util` subdirectory.

collinss-jpl (changes requested):
Everything looks good, this branch is very close to being merge ready, but the scripts in `py_util` are actually all PGE specific (despite the generic file names), and need to be moved to the appropriate PGE-specific directories to accomplish the goal of this ticket. Once they're moved (and any affected integration test scripts are updated to reference the new locations of the comparison scripts) this branch should be good to go.

.ci/scripts/dswx_hls/build_dswx_hls.sh
JimHofman: Fixed

collinss-jpl (approved):

Issue #393, opened 2024-01-17
[New Feature]: Improve error reporting for SAS subprocess failures


collinss-jpl:
### Checked for duplicates

Yes - I've already checked

### Alternatives considered

Yes - and alternatives don't suffice

### Related problems

When [executing the SAS as a subprocess within the PGE](https://github.com/nasa/opera-sds-pge/blob/main/src/opera/util/run_utils.py#L203), if the SAS fails or crashes for any reason, the specific details of the error are only captured to the combined PGE/SAS log. [The exception handling code merely reports that the SAS subprocess exited with a non-zero return code](https://github.com/nasa/opera-sds-pge/blob/main/src/opera/util/run_utils.py#L211), or similar, and this is what gets reported directly to the SDS Figaro web console that operators interact with.

Having to dig into the logs to derive the true error is not ideal and can lead to confusion on behalf of the operators in terms of what they track in their own error reporting tools.

### Describe the feature request

The goal of this ticket is to enhance the exception handling logic for a failed SAS subprocess to try to extract whatever error info or stack trace might be within the captured log, and include it with the message that is logged to the console. 

This could be as simple as always grabbing the last ~25 lines or so of the log (similar to running `tail -n25` on the log file), or we could do some more sophisticated parsing to find and extract only the stack trace using regular expressions or similar.

collinss-jpl:
Closed by #407

PR #407, opened 2024-02-29
393 improve error reporting for sas subprocess

rjbdev05:
## Description
- When executing the SAS as a subprocess, if SAS fails, the specific details of error message is only captured in log which is deleted at the end of the processing. This branch fixes so that the output to console gives the last 20 lines of the tracebaak of the error messages.

## Affected Issues
- #393

## Testing
- In the directory, opera-sds-pge/src/opera/test/pge/base/ 
- python3 test_base_pge.py 2>&1 | tee ../../../../test_base_pge_error.log
- <Once the PR is created, don't forget to check the Build/Test Jenkins pipeline for your branch to ensure all tests are passing before merge>


src/opera/__pycache__/__init__.cpython-310.pyc
collinss-jpl, line 1: All of these `__pycache__` directories that have been checked in need to be removed.

src/opera/test/data/test_dswx_s1_config_error.yaml
```diff
@@ -0,0 +1,65 @@
...
+          shoreline_shapefile: dswx_s1_pge_test/input_dir/shoreline.shp
+      ProductPathGroup:
+        OutputProductPath: dswx_s1_pge_test/output_dir
+        ScratchPath: dswx_s1_pge_test/scratch_dir
+      PrimaryExecutable:
+        ProductIdentifier: DSWX_S1
+        ProductVersion: "1.0"
+        ProgramPath: /bin/echo
+        ProgramOptions:
+        - "There's been some sprort of error"
```
collinss-jpl, line 25: "sprort" -> "sort"

src/opera/util/logger.py
```diff
@@ -81,6 +81,7 @@ def default_log_file_name():
     log_datetime_str = time_util.get_time_for_filename(datetime.datetime.now())
     file_path = f"pge_{log_datetime_str}.log"
 
+    print ("file_path = ",file_path)
```
collinss-jpl, line 84: Let's remove this line before the branch is merged

src/opera/util/run_utils.py
```diff
@@ -207,10 +207,17 @@ def time_and_execute(command_line, logger, execute_via_shell=False):
...
 
+    tail = run_result.stdout.decode().split('\n')
+    tail = "\n".join(tail[-20:])
+
     if run_result.returncode:
         error_msg = (f'Command "{" ".join(command_line)}" failed with exit '
-                     f'code {run_result.returncode}')
+                f'code {run_result.returncode}, traceback: {tail}')
 
+#    if run_result.returncode:
```
collinss-jpl, line 217: This commented out block can also be removed

src/opera/util/run_utils_error.py
collinss-jpl, line 1: This file should be removed before merge

collinss-jpl (changes requested):
The core changes look good, but there's a lot that should be trimmed from this branch before it gets merged. I am also waiting on SA support to update some expired credentials in our Jenkins system, which is why the current build fails. Once the credentials are updated, I will re-run this branch to ensure the tests are passing before merging.

collinss-jpl:
Went ahead and addressed my findings and got the branch passing the CI system, will merge shortly

collinss-jpl (approved):

PR #427, opened 2024-04-11
416 refactor cscl s1 int test

JimHofman:
## Description
- Refactoring the cslc_s1 int shell scripts for baseline and static workflows
    - Required test files now coming from S3
    - 2 docker runs, baseline and static

## Testing
run the test_int_cscl_s1.sh script on one of our aws machines
run Jenkins and verify results


collinss-jpl (approved):
Confirmed that int test for CSLC-S1 runs to completion and all comparisons pass! Thanks @JimHofman!

JimHofman (commented):
Looks good.

PR #441, opened 2024-05-16
Enhanced traceback capture for SAS failures

### LINKED ISSUE #423: [New Feature]: Enhanced Traceback capture for failed SAS subprocess
### Checked for duplicates

Yes - I've already checked

### Alternatives considered

Yes - and alternatives don't suffice

### Related problems

This feature is to enhance the error reporting performed when a SAS subprocess fails. Currently, we simply grab the last 20 or so lines from the captured log output of the SAS subprocess, and include it in the error message provided to the end-user. However, this is not ideal, as important parts of the stack trace could potentially be truncated from the final error message if the Traceback is longer than 20 lines. Conversely, we could also include unrelated lines if the Traceback is shorter than 20 lines.

### Describe the feature request

The goal of this ticket would be to add additional sophistication to the parsing of the captured SAS output such that the entire Traceback stack is included with the error message. This can likely be accomplished by utilizing a regular expression over the captured log output to fully capture the Traceback.

A cursory google search for "python traceback regex" turned up the following solutions from stackexchange that look promising:

https://stackoverflow.com/a/45853331
https://stackoverflow.com/a/53658873
https://stackoverflow.com/a/70827258

Any or all of these examples can be incorporated into the implementation for this feature, as long as it accomplishes the stated goal of the ticket (i.e. only provide a complete Traceback stack in an error message for a failed SAS subprocess).

collinss-jpl:
## Description
- This branch enhances the error reporting from SAS failures by utilizing a regular expression to capture a full traceback stack from the log, rather than just the last 20 lines.

## Affected Issues
- Resolves #423

## Testing
- Unit test for a failing SAS condition has been updated to ensure the full traceback (which is longer than 20 lines in this test) is now provided in the error message that would be reported to an SDS end user.

### DIFF
--- a/src/opera/test/data/test_sas_error_config.yaml
+++ b/src/opera/test/data/test_sas_error_config.yaml
@@ -26,42 +26,38 @@ RunConfig:
         ProgramOptions:
           - -c
           - echo START;
-          - echo sample traceback line 2;
-          - echo Traceback \(most recent call last):;
-          - echo File "/usr/lib/python2.7/runpy.py", line 174, in _run_module_as_main;
-          - echo "__main__", fname, loader, pkg_name);
-          - echo File "/usr/lib/python2.7/runpy.py", line 72, in _run_code;
-          - echo exec code in run_globals;
-          - echo File "/root/env/common/test/test/__main__.py", line 5, in <module>;
-          - echo main();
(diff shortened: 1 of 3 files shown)


JimHofman (commented):
Looks good.

PR #451, opened 2024-06-05
439 dswx ni filename validation

JimHofman:
## Description
- Added file name validation based on a regular expression

## Affected Issues
N/A

## Testing
Added a unit test

### DIFF
--- a/src/opera/pge/dswx_ni/dswx_ni_pge.py
+++ b/src/opera/pge/dswx_ni/dswx_ni_pge.py
@@ -10,7 +10,7 @@ from NISAR (NI) PGE.
 """
 
 import re
-from os.path import join
+from os.path import basename, join
 
 from opera.pge.base.base_pge import PgeExecutor
 from opera.pge.dswx_s1.dswx_s1_pge import DSWxS1PostProcessorMixin, DSWxS1PreProcessorMixin
@@ -63,10 +63,16 @@ class DSWxNIPostProcessorMixin(DSWxS1PostProcessorMixin):
 
     def _validate_output_product_filenames(self):
         """
-        Test method to verify the regular expression used to
-        validate the output product file names assigned by the SAS
-        This method will validate that the filename has the acceptable name
-        through a regular expression.  If the pattern does not match
+        This method validates output product file names assigned by the SAS
+        via a regular expression.  The output product file names should follow
+        this conventions:
+
+         <PROJECT>_<LEVEL>_<PRODUCT TYPE>_<SOURCE>_<TILE ID>_<ACQUISITION TIMESTAMP>_
+         <CREATION TIMESTAMP>_<SENSOR>_<SPACING>_<PRODUCT VERSION>_<BAND INDEX>_
+         <BAND NAME>_<FILE EXTENSION>
+
+         If the pattern does not match a critical error will cause a RuntimeError.
+
         """
         validated_product_filenames = []
         pattern = re.compile(
@@ -76,9 +82,7 @@ class DSWxNIPostProcessorMixin(DSWxS1PostProcessorMixin):
             r'(?P<band_name>WTR|BWTR|CONF|DIAG)|_BROWSE)?[.](?P<ext>tif|tiff|png)$')
 
         for output_file in self.runconfig.get_output_product_filenames():
-            if pattern.match(output_file.split('/')[-1]):
-                validated_product_filenames.append(output_file)
-            else:
+            if not pattern.match(basename(output_file)):
                 error_msg = (f"Output file {output_file} does not match the output predict "
                              f"naming convention.")
                 self.logger.critical(self.name, ErrorCode.INVALID_OUTPUT, error_msg)
--- a/src/opera/pge/dswx_ni/dswx_ni_pge.py
+++ b/src/opera/pge/dswx_ni/dswx_ni_pge.py
@@ -10,11 +10,10 @@ from NISAR (NI) PGE.
 """
 
 import re
-
 from os.path import join
 
 from opera.pge.base.base_pge import PgeExecutor
-from opera.pge.dswx_s1.dswx_s1_pge import DSWxS1PreProcessorMixin, DSWxS1PostProcessorMixin
+from opera.pge.dswx_s1.dswx_s1_pge import DSWxS1PostProcessorMixin, DSWxS1PreProcessorMixin
 from opera.util.error_codes import ErrorCode
 from opera.util.time import get_time_for_filename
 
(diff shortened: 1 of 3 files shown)


src/opera/pge/dswx_ni/dswx_ni_pge.py
```diff
@@ -60,6 +61,28 @@ class DSWxNIPostProcessorMixin(DSWxS1PostProcessorMixin):
...
+        validated_product_filenames = []
+        pattern = re.compile(
+            r'(?P<project>OPERA)_(?P<level>L3)_(?P<product_type>DSWx)-(?P<source>NI)_(?P<tile_id>T[^\W_]{5})_'
+            r'(?P<acquisition_ts>\d{8}T\d{6}Z)_(?P<creation_ts>\d{8}T\d{6}Z)_(?P<sensor>LSAR)_(?P<spacing>30)_'
+            r'(?P<product_version>v\d+[.]\d+)(_(?P<band_index>B\d{2})_'
+            r'(?P<band_name>WTR|BWTR|CONF|DIAG)|_BROWSE)?[.](?P<ext>tif|tiff|png)$')
+
+        for output_file in self.runconfig.get_output_product_filenames():
+            if pattern.match(output_file.split('/')[-1]):
+                validated_product_filenames.append(output_file)
```
collinss-jpl, line 80: I don't think its necessary to compile a list of the validated file names. Since we'll throw an exception if any one file name fails, the list will always end up being equal to the set of all filenames anyhow. So we can simplify to a single conditional like so:

```python
if not pattern.match(os.path.basename(output_file)):  # basename is more intuitive to read than .split('/')[-1]
    error_msg = (...)
    self.logger.critical(self.name, ErrorCode.INVALID_OUTPUT, error_msg)
```

src/opera/test/pge/dswx_ni/test_dswx_ni_pge.py
```diff
@@ -119,6 +119,7 @@ def generate_band_data_output(self, band_data, empty_file=False, clear=True):
             else:
                 os.system(f"touch {join(self.test_output_dir, band_output_file)}")
 
+    # @unittest.skip("SKIP")
```
collinss-jpl, line 122: Let's remove this before merge

src/opera/test/pge/dswx_ni/test_dswx_ni_pge.py
```diff
@@ -389,6 +390,90 @@ def test_dswx_ni_pge_output_validation(self):
...
+
+        with open(runconfig_path, 'r', encoding='utf-8') as stream:
+            runconfig_dict = yaml.safe_load(stream)
+
+        with open(runconfig_path, 'w', encoding='utf-8') as outfile:
+            yaml.safe_dump(runconfig_dict, outfile, sort_keys=False)
+
+        pge = DSWxNIExecutor(pge_name="DSWxNIPgeTest", runconfig_path=runconfig_path)
+
+        pge.run_preprocessor()
```
collinss-jpl, line 406: Let's add a comment here that we run only the preprocessor to parse and ingest the RunConfig

src/opera/test/pge/dswx_ni/test_dswx_ni_pge.py
```diff
@@ -389,6 +390,90 @@ def test_dswx_ni_pge_output_validation(self):
...
+        self.generate_band_data_output(band_data, clear=True)
+
+        pge = DSWxNIExecutor(pge_name="DSWxNIPgeTest", runconfig_path=runconfig_path)
+
+        pge.run_preprocessor()
+
+        with self.assertRaises(RuntimeError):
+            pge._validate_output_product_filenames()
+
+        # Note: the log files exists because the critical() method in the PgeLogger class save it to disk.
```
collinss-jpl, line 443: 👍 
Forgot that it even works this way!

collinss-jpl (changes requested):
Overall it looks good, just a little bit of cleanup to be done on the new `_validate_output_product_filenames` function.

src/opera/test/pge/dswx_ni/test_dswx_ni_pge.py
JimHofman: oops

src/opera/test/pge/dswx_ni/test_dswx_ni_pge.py
JimHofman: Done

src/opera/pge/dswx_ni/dswx_ni_pge.py
```diff
@@ -60,6 +61,32 @@ class DSWxNIPostProcessorMixin(DSWxS1PostProcessorMixin):
...
 
+    def _validate_output_product_filenames(self):
+        """
+        This method validates output product file names assigned by the SAS
+        via a regular expression.  The output product file names should follow
+        this conventions:
+
+         <PROJECT>_<LEVEL>_<PRODUCT TYPE>_<SOURCE>_<TILE ID>_<ACQUISITION TIMESTAMP>_
+         <CREATION TIMESTAMP>_<SENSOR>_<SPACING>_<PRODUCT VERSION>_<BAND INDEX>_
+         <BAND NAME>_<FILE EXTENSION>
```
collinss-jpl, line 72: Should be `<BAND NAME>.<FILE EXTENSION>`

src/opera/pge/dswx_ni/dswx_ni_pge.py
```diff
@@ -60,6 +61,32 @@ class DSWxNIPostProcessorMixin(DSWxS1PostProcessorMixin):
...
+        validated_product_filenames = []
+        pattern = re.compile(
+            r'(?P<project>OPERA)_(?P<level>L3)_(?P<product_type>DSWx)-(?P<source>NI)_(?P<tile_id>T[^\W_]{5})_'
+            r'(?P<acquisition_ts>\d{8}T\d{6}Z)_(?P<creation_ts>\d{8}T\d{6}Z)_(?P<sensor>LSAR)_(?P<spacing>30)_'
+            r'(?P<product_version>v\d+[.]\d+)(_(?P<band_index>B\d{2})_'
+            r'(?P<band_name>WTR|BWTR|CONF|DIAG)|_BROWSE)?[.](?P<ext>tif|tiff|png)$')
+
+        for output_file in self.runconfig.get_output_product_filenames():
+            if not pattern.match(basename(output_file)):
+                error_msg = (f"Output file {output_file} does not match the output predict "
```
collinss-jpl, line 86: I think "predict" can be removed here

src/opera/pge/dswx_ni/dswx_ni_pge.py
JimHofman: good catch. Fixed

src/opera/pge/dswx_ni/dswx_ni_pge.py
JimHofman: Done

collinss-jpl (approved):

PR #453, opened 2024-06-11
flake8 cleanup

JimHofman:
## Description
Ticket 435:  Flake8 cleanup for R4

## Affected Issues
N/A

## Testing
Ran pytest for all modules.  All passed 93% total coverage.  A few tests need a little work to get above 90%.


collinss-jpl (changes requested):
Hi @JimHofman, changes look good. Running flake8 shows no remaining findings for the code within the `src` directory. However, when I run `flake8` from _the top level of the repository_, we get a number of findings for the code within the `.ci` directory (the python product comparison scripts in particular).

Here is a log of findings that I created based off of this branch: [flake8.log](https://github.com/user-attachments/files/15808285/flake8.log)

Most of the findings seem pretty trivial to address, so if you wouldn't mind knocking them out as well it would be much appreciated!

JimHofman:
Thanks Scott.  I'm pretty sure I got them all this time.  There was a lot of changes, but it's good to have that compare code that we did not write at least looking somewhat similar to our code.  I don't mind doing this since I thought it up and it's a lot of busy work.  But I think it might make the code more consistent and cleaner if we modified the docstrings  further on the scripts in .ci, and have them match our style.

PR #494, opened 2024-08-21
Remove 'python -m' capabillity from run_utils.py

JimHofman:
## Description
- Removed a condition that called for using 'python -m' and replaced the clause with an error message.
- Made changes to run_utils.py, base_pge.py and their unit tests
- Also made cosmetic chages to test_dswx_ni_pge.py and test_dsws_s1_pge.py after running flake8 on the test/ directory.

## Affected Issues
-  None
## Testing
- Ran unit tests

### DIFF
--- a/src/opera/util/run_utils.py
+++ b/src/opera/util/run_utils.py
@@ -84,7 +84,7 @@ def get_traceback_from_log(log_contents):
     return trackback_match
 
 
-def create_sas_command_line(logger, pge_name, sas_program_path, sas_runconfig_path,
+def create_sas_command_line(sas_program_path, sas_runconfig_path,
                             sas_program_options=None):
     """
     Forms the appropriate command line for executing a SAS from the parameters
@@ -97,10 +97,6 @@ def create_sas_command_line(logger, pge_name, sas_program_path, sas_runconfig_pa
 
     Parameters
     ----------
-    logger :  PGELogger
-        The PGELogger instance
-    pge_name : str
-        PGE name
     sas_program_path : str
         The path to the SAS executable to be invoked by the returned command line.
     sas_runconfig_path : str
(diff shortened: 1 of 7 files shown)


RKuttruff (approved):
Looks good to me

src/opera/util/run_utils.py
```diff
@@ -131,9 +135,9 @@ def create_sas_command_line(sas_program_path, sas_runconfig_path,
             if not os.access(executable_path, mode=os.X_OK):
                 raise OSError(f"Requested SAS program path {sas_program_path} exists, "
                               f"but does not have execute permissions.")
-        # Otherwise, sas_program_path might be a python module path
         else:
-            command_line = ['python3', '-m', sas_program_path]
+            error_msg = "Configured executable from the RunConfig could not be found"
```
collinss-jpl, line 139: Similar to the error message above, lets include `{sas_program_path}` somewhere in this message to assist with troubleshooting.

Same applies for the error message in `create_qa_command_line`

src/opera/util/run_utils.py
```diff
@@ -131,9 +135,9 @@ def create_sas_command_line(sas_program_path, sas_runconfig_path,
...
+            logger.critical(pge_name, ErrorCode.EXECUTABLE_FROM_RUNCONFIG_NOT_FOUND, error_msg)
```
collinss-jpl, line 140: Rather than make the logger.critical call within this function (which necessitates passing a logger in as an additional argument) I think I would rather have this function raise an `OSError` for this case.

That way, the callers of the function (the PGE classes) will catch the OSError, then use their own logger instance to make the `critical` call.

JimHofman: Yeah, I saw the back to back errors reporting.

src/opera/test/test
collinss-jpl, line 1: Please remove this file before we merge this branch

PR #500, opened 2024-08-30
Reworked generation of custom metadata in DSWx-S1 and DSWx-NI to utilize the granule filename instead of the core filename

RKuttruff:
Scott's patch for DSWx-S1 ISO XML issue


RKuttruff:
Scott's changes look good to me. (I can't approve since I opened the PR)
Will wait for checks to complete

PR #505, opened 2024-09-04
ticket 489 - metadata: log error message if no file is found

JimHofman:
## Description
- Add code to the PGE's to test for the existence of a product file that yields metadata when collected metadata

## Affected Issues
- N/A

## Testing
- Updated 5 unit tests to 
- Added code but did not add test code to  pge's (see comments below)


RKuttruff (approved):
This looks good

src/opera/pge/cslc_s1/cslc_s1_pge.py
```diff
@@ -619,7 +619,12 @@ def _collect_cslc_product_metadata(self, metadata_product):
             use with the ISO metadata Jinja2 template.
 
         """
-        output_product_metadata = get_cslc_s1_product_metadata(metadata_product)
+        # Extract all metadata assigned by the SAS at product creation time
+        try:
+            output_product_metadata = get_cslc_s1_product_metadata(cslc_product)
+        except Exception as err:
+            msg = f'Could not extract metadata from {cslc_product}: Exception {err}'
```
collinss-jpl, line 626: This is nit-picky for sure, but for the sake of uniformity I would like these error messages to be formatted in a way that aligns with other logged errors. Something like:

```
msg = f'Failed to extract metadata from {cslc_product}, reason: {err}'
```

collinss-jpl (commented):
Changes look good, particularly the unit test updates. I only have one request regarding the format of the logged error messages.

JimHofman:
Scott I'm glad you made this request because I realized I did not cover the new code in a couple of unit tests.

PR #506, opened 2024-09-05
494 - delete condition that might trigger erroneous error in input_validatio…

JimHofman:
## Description
- Remove problematic 'if' block that might trigger an 'intermittent'/'conditional' error. 

## Affected Issues
- N/A

## Testing
- No specific test, but ran the normal pytest to test everything.


RKuttruff (commented):
So if I'm understanding correctly, the check here would fail the PGE if the runconfig contained duplicate RTC burst IDs? What would be the behavior if there were duplicates in the runconfig or are they deduplicated elsewhere?

JimHofman:
Hi Riley.  We are testing input CSLC burst ID's can have duplicate burst IDs because they are products of both S1A and S1B.  So in that case we would get an erroneous critical error.

collinss-jpl (approved):

Issue #514, opened 2024-09-23
Enhancements for ISO Measured Parameters Description Schema


collinss-jpl:
As a final refinement to the automated Measured Parameter population feature for ISO XML products, I would like to see the following:

- [x] Refactor the inline schema for the Parameters Description YAML file into its own file. This file should live under `src/pge/base/schema` since it will be common to all PGEs.
- [x] Add code to `base_pge.py` to load the schema from file before validating the Parameters Description file via Yamale
- [x] Enhance the Descriptions schema file with the following additional fields:
  - An optional "display" (display_name? display_override?) field, that when provided, takes precedence over the default behavior used to "canonicalize" the parameter name. This would replace the current behavior of repeated `replace()` calls to adjust things like acronyms in parameter names
  - A required "attribute_type" field which we can use to capture the values used for the `EOS_AdditionalAttributeTypeCode` tags in the original ISO templates. This field should be an enumeration and the [allowable values](https://git.earthdata.nasa.gov/projects/EMFD/repos/iso-schemas/browse/resources/Codelist/eosCodelists.xml?at=606229ea84f207f42b7b477c507eb422f8996115#23) should match those in the `eosCodelists.xml` schema used to validate ISO products.
- [x] Update the existing Parameters Description config for DSWx-S1 to add the new fields. For the attribute_type field, reference the original values used in the current DSWx-S1 ISO template currently in the `main` branch.

RKuttruff:
Closed by #516

PR #516, opened 2024-09-25
ISO XML Measured Parameter Section Further Enhancements

RKuttruff:
## Description
- Moved inline Measured Parameters Config (MPC) schema into a dedicated file `src/opera/pge/base/schema/iso_metadata_measured_parameters_config_schema.yaml`
- Added fields to MPC schema:
  - `attribute_type`: The parameter's EOS AdditionalAttributeDataTypeCode ([info](https://git.earthdata.nasa.gov/projects/EMFD/repos/iso-schemas/browse/resources/Codelist/eosCodelists.xml?at=606229ea84f207f42b7b477c507eb422f8996115#23)) [REQUIRED]
    - Must be one of: `geographicIdentifier`, `qualityInformation`, `instrumentInformation`, `sensorInformation`, `contentInformation`, `platformInformation`, `citation.identifier`, `descriptiveKeyword`, `processingInformation`, `processingParameter`, or `commandLineArgument`
  - `display_name`: Override for parameter's display name that is normally derived from the metadata variable name. [OPTIONAL]
  - `attribute_data_type`: Override for parameter's data type that is normally guessed from metadata variable's Python `type`. ([info](https://git.earthdata.nasa.gov/projects/EMFD/repos/iso-schemas/browse/resources/Codelist/eosCodelists.xml?at=606229ea84f207f42b7b477c507eb422f8996115#105)) [OPTIONAL]
    - Must be one of: `string`, `float`, `int`, `boolean`, `date`, `time`, `dateTime`, `dateString`, `timeString`, or `dateTimeString`
- Updated DSWx-S1 PGE to use the new MPC fields to populate the ISO XML template, falling back to guesses for absent, non-required fields.

## Affected Issues
- #514 

## Testing
- Manually ran unit tests, injecting a copy statement to pull out the temporary ISO XML files for [diffing](https://github.com/user-attachments/files/17156313/514-diff.html.zip).
```python
# in test_dswx_s1_pge.py @ line 290
shutil.copy(expected_iso_metadata_file, '/tmp/.')
```
- Ran int test on dev-pge, checking the produced XML files

### DIFF
--- a/src/opera/pge/base/schema/iso_metadata_measured_parameters_config_schema.yaml
+++ b/src/opera/pge/base/schema/iso_metadata_measured_parameters_config_schema.yaml
@@ -1,7 +1,22 @@
 map(include('parameter'), key=str())
 ---
 parameter:
+
+  # REQUIRED: Plain-text description of the Measured Parameter as provided in the product spec document
   description: str()
+
+  # REQUIRED: NASA/EOSDIS additional attribute type.
+  # Valid values defined here: https://git.earthdata.nasa.gov/projects/EMFD/repos/iso-schemas/browse/resources/Codelist/eosCodelists.xml?at=606229ea84f207f42b7b477c507eb422f8996115#23
   attribute_type: enum('geographicIdentifier', 'qualityInformation', 'instrumentInformation', 'sensorInformation', 'contentInformation', 'platformInformation', 'citation.identifier', 'descriptiveKeyword', 'processingInformation', 'processingParameter', 'commandLineArgument')
-  attribute_name: str(required=False)
+
+  # OPTIONAL: The data type of the attribute. The PGE will attempt to guess this based on the returned Python datatype
+  # provided by the TIFF metadata util, but this may not always be correct (ie: version strings being interpreted as
+  # a float); use this field to correct errors where needed.
+  # Valid values defined here: https://git.earthdata.nasa.gov/projects/EMFD/repos/iso-schemas/browse/resources/Codelist/eosCodelists.xml?at=606229ea84f207f42b7b477c507eb422f8996115#105
   attribute_data_type: enum('string', 'float', 'int', 'boolean', 'date', 'time', 'dateTime', 'dateString', 'timeString', 'dateTimeString', required=False)
+
+  # OPTIONAL: The display name of the attribute. The PGE will attempt to derive this from the internal metadata variable
+  # name by converting to title case and stripping out underscores. This should provide the proper display name, but may
+  # handle some acronyms improperly (ie, DSWx, MGRS, RTC) and some desired display names retain some underscores. Use
+  # this field to correct these errors where needed.
+  display_name: str(required=False)
--- a/src/opera/pge/base/schema/iso_metadata_measured_parameters_config_schema.yaml
+++ b/src/opera/pge/base/schema/iso_metadata_measured_parameters_config_schema.yaml
@@ -1,0 +1,7 @@
+map(include('parameter'), key=str())
+---
+parameter:
+  description: str()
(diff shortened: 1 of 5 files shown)


src/opera/pge/base/schema/iso_metadata_measured_parameters_config_schema.yaml
```diff
@@ -0,0 +1,7 @@
+map(include('parameter'), key=str())
+---
+parameter:
+  description: str()
+  attribute_type: enum('geographicIdentifier', 'qualityInformation', 'instrumentInformation', 'sensorInformation', 'contentInformation', 'platformInformation', 'citation.identifier', 'descriptiveKeyword', 'processingInformation', 'processingParameter', 'commandLineArgument')
+  attribute_name: str(required=False)
```
collinss-jpl, line 6: For the sake of making the purpose of this field a little clearer, lets call it `display_name`

src/opera/pge/base/schema/iso_metadata_measured_parameters_config_schema.yaml
```diff
@@ -0,0 +1,7 @@
+map(include('parameter'), key=str())
+---
+parameter:
```
collinss-jpl, line 3: One nice thing about yaml schemas is that, unlike JSON, you can provide inline documentation. It would be nice to see each of these parameters documented with things like what the fallback behavior is when an optional field is omitted. We can also add links to the ISO schema to justify why the enums are defined as they are.

src/opera/pge/dswx_s1/templates/dswx_s1_measured_parameters.yaml
```diff
@@ -1,134 +1,217 @@
...
 PROCESSING_INFORMATION_FILTER:
   description: RTC intensity filtering method (bregman, lee, guided_filter, anisotropic_diffusion, etc.).
+  attribute_type: processingParameter
 PROCESSING_INFORMATION_FILTER_ENABLED:
   description: Indicates if the filter is applied to the input RTC.
+  attribute_type: processingParameter
 PROCESSING_INFORMATION_FILTER_OPTION:
   description: Parameters used for the despeckle filter.
+  attribute_type: processingParameter
+  attribute_name: ProcessingInformationFilterOptions
```
collinss-jpl, line 63: I take it this was the name that used in the original template, but I could see someone complaining that there is a mismatch between the two names so lets just omit this going forward.

src/opera/pge/dswx_s1/dswx_s1_pge.py
```diff
@@ -458,19 +458,17 @@ def _augment_measured_parameters(self, measured_parameters):
             if isinstance(value, list):
                 value = json.dumps(value)
 
-            data_type = python_type_to_xml_type(value)
-
-            # TODO: This is hardcoded for now, but we should eventually try to guess this or add it to the descriptions
-            #  YAML file
-            attr_type = 'processingInformation'
+            guessed_data_type = python_type_to_xml_type(value)
+            guessed_attr_name = (name.title()
```
collinss-jpl, line 462: Couple of things I would like to see here: lets lose the `.replace('Mgrs', 'MGRS').replace('Dswx', 'DSWx')` in favor of the new display name override field in the config for any of the parameters that include 'DSWx' or 'MGRS'.

For the remaining transformation (name.title().replace('_', '')`, lets factor this out into a simple utility function that we can reuse when we roll out this feature to the other PGEs.

collinss-jpl (changes requested):

src/opera/pge/dswx_s1/dswx_s1_pge.py
RKuttruff: Where would I put this util function?

Somewhere under `src/opera/util/` perhaps but in which module?

collinss-jpl: Since its related to rendering of the metadata, `util/render_jinja2.py` is as good a spot as any

collinss-jpl (approved):
Changes look perfect, thanks again @RKuttruff!

PR #532, opened 2024-10-14
.secrets.baseline update procedure improvements

### LINKED ISSUE #518: Improve instructions for updating .secrets.baseline reported by CI system
A recent branch triggered the logic that detects an update to the .secrets.baseline file and subsequently fails the build/test job in the CI system. This is the expected behavior, but the instructions provided to the developer on how to update .secrets.baseline to pass the check could use some improvements.

The goal of this ticket is to make the following improvements to the instructions in `.ci/scripts/util/secrets_scan_compare.sh`:

- [x] Remove "You can find detailed instructions at this link: https://nasa-ammos.github.io/slim/continuous-testing/starter-kits/#detect-secrets" from line 26. We'll put the appropriate instructions directly into this script (the link seems to be dead anyhow)
- [x] Add an additional line after line 27 with the appropriate command to run: `detect-secrets scan --all-files --disable-plugin AbsolutePathDetectorExperimental --exclude-files '\.secrets.*' --exclude-files '\.git*' --exclude-files 'test_results' --exclude-files '\.pytest_cache' --exclude-files '\.venv' --exclude-files 'venv' --exclude-files 'dist' --exclude-files 'build' --exclude-files '.*\.egg-info' > .secrets.baseline`
- [x] Add an additional step after this with the instructions for running the "audit" mode: `detect-secrets audit .secrets.baseline`. Be sure to add a detail that if a detected "secret" is benign, then "y" should be entered as the answer to "Is this a secret that should be committed to this repository?" to clear up any ambiguity here.

This should hopefully streamline the process a bit since we will likely trip this check multiple times as we roll out the ISO XML template updates to each PGE

collinss-jpl:
## Description
- This branch makes some small adjustments to the instructions provided by the CI system when the secrets detection step results in new findings within the repo. The commands to be run are now explicitly described, and a step for running the "audit' phase of updating .secrets.baseline has been added.
- There's also a fix for an unescaped character in secrets_scan.sh

## Affected Issues
- Resolves #518 

## Testing
- N/A

### DIFF
--- a/.ci/scripts/util/secrets_scan_compare.sh
+++ b/.ci/scripts/util/secrets_scan_compare.sh
@@ -23,9 +23,19 @@ then
     echo "" >&2
     echo "Please follow the steps below on your local machine to reveal and handle the secrets:" >&2
     echo "" >&2
-    echo "1️⃣ Run the 'detect-secrets' tool on your local machine. This tool will identify and clean up the secrets. You can find detailed instructions at this link: https://nasa-ammos.github.io/slim/continuous-testing/starter-kits/#detect-secrets" >&2
+    echo "1️⃣ Run the 'detect-secrets' tool on your local machine. This tool will identify and clean up the secrets:" >&2
     echo "" >&2
-    echo "2️⃣ After cleaning up the secrets, commit your changes and re-push your update to the repository." >&2
+    echo "    \$ detect-secrets scan --all-files --disable-plugin AbsolutePathDetectorExperimental --exclude-files '\.secrets\..*' --exclude-files '\.git.*' --exclude-files 'test_results' --exclude-files '\.pytest_cache' --exclude-files '\.venv' --exclude-files 'venv' --exclude-files 'dist' --exclude-files 'build' --exclude-files '.*\.egg-info' > .secrets.baseline" >&2
+    echo "" >&2
+    echo "2️⃣ Perform an audit on the updated .secrets.baseline to disposition any new findings:" >&2
+    echo "" >&2
+    echo "    \$ detect-secrets audit .secrets.baseline" >&2
+    echo "" >&2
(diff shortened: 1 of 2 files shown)


.ci/scripts/util/secrets_scan_compare.sh
```diff
@@ -23,9 +23,19 @@ then
...
-    echo "2️⃣ After cleaning up the secrets, commit your changes and re-push your update to the repository." >&2
+    echo "    \$ detect-secrets scan --all-files --disable-plugin AbsolutePathDetectorExperimental --exclude-files '\.secrets\..*' --exclude-files '\.git.*' --exclude-files 'test_results' --exclude-files '\.pytest_cache' --exclude-files '\.venv' --exclude-files 'venv' --exclude-files 'dist' --exclude-files 'build' --exclude-files '.*\.egg-info' > .secrets.baseline" >&2
+    echo "" >&2
+    echo "2. Perform an audit on the updated .secrets.baseline to disposition any new findings:" >&2
+    echo "" >&2
+    echo "    \$ detect-secrets audit .secrets.baseline" >&2
+    echo "" >&2
+    echo "The audit tool will ask the following for all newly detected \"secrets\": \"Is this a secret that should be committed to this repository?\"" >&2
+    echo "" >&2
+    echo "If the detected secret is benign (i.e. a public email address or dummy password) the correct answer is \"y\"." >&2
```
RKuttruff, line 36: Maybe there should be a little more attention called out to this line considering how counterintuitive it is?

RKuttruff (approved):
All looks good, I just had one comment on the message

.ci/scripts/util/secrets_scan_compare.sh
RKuttruff: I'd imagine you can just put the emoji characters directly into the .sh file 
I just tried this in PyCharm and it worked no problem

Barring that, you could try a plain text approach like even a `NOTE: ` before the message.

Perhaps move the note to before the step with the audit command so users are aware that `'y' == safe` _before_ running the command

Issue #524, opened 2024-09-30
ISO XML Measured Parameters Update for DISP-S1


collinss-jpl:
- [x] Creation of the Measured Parameters Description YAML config 
  - Latest product spec can be found in Artifactory under the following path: `general/gov/nasa/jpl/opera/adt/disp_s1/r5/calval/documents/OPERA_DISP_S1_Calval_Product_Spec.docx`
- [x] Update the existing ISO XML template to use a [jinja2 for-loop construct](https://github.com/nasa/opera-sds-pge/blob/main/src/opera/pge/dswx_s1/templates/OPERA_ISO_metadata_L3_DSWx_S1_template.xml.jinja2#L586-L608) for the MeasuredParameters section. This should be performed after creation of the Parameter Description file described above

- [x] Update `_collect_disp_s1_product_metadata()` to include a call to augment_measured_parameters() to ensure all [metadata is formatted as expected](https://github.com/nasa/opera-sds-pge/blob/b5cb72869d440247e0e5e9a3fec0d25840abefdc/src/opera/pge/dswx_s1/dswx_s1_pge.py#L448) before the template is instantiated

- [x] After all changes are made, run the integration test for the PGE (either via Jenkins or locally on dev machine) and inspect the resulting ISO XML product to ensure `MeasuredParameters` section is filled out as expected.

Issue #520, opened 2024-09-30
Roll out ISO XML Measured Parameter Generation feature to remaining PGEs


collinss-jpl:
Now that the new feature for automated population of Measured Parameters within ISO XML has been developed and tested with DSWx-S1, we need to roll the feature out the remaining PGEs. This should entail the following changes:

- Creation of the Measured Parameters Description YAML config (see the [DSWx-S1 version](https://github.com/nasa/opera-sds-pge/blob/main/src/opera/pge/dswx_s1/templates/dswx_s1_measured_parameters.yaml) for an example). Note this task will also require the latest Product Spec document for the SAS to get the appropriate descriptions, as well as the existing ISO XML template to pull the appropriate values for `attribute_type`
- Update the existing ISO XML template to use a [jinja2 for-loop construct](https://github.com/nasa/opera-sds-pge/blob/main/src/opera/pge/dswx_s1/templates/OPERA_ISO_metadata_L3_DSWx_S1_template.xml.jinja2#L586-L608) for the MeasuredParameters section. This should be performed after creation of the Parameter Description file described above
- Update the PGE's implementation of `_collect_<pge>_product_metadata()` to include a call to `augment_measured_parameters()` to ensure all [metadata is formatted as expected](https://github.com/nasa/opera-sds-pge/blob/b5cb72869d440247e0e5e9a3fec0d25840abefdc/src/opera/pge/dswx_s1/dswx_s1_pge.py#L448) before the template is instantiated
- After all changes are made, run the integration test for the PGE (either via Jenkins or locally on dev machine) and inspect the resulting ISO XML product to ensure `MeasuredParameters` section is filled out as expected.

- [x] #521
- [x] #522
- [x] #523
- [x] #524

RKuttruff:
Hi @collinss-jpl and @JimHofman, yesterday I played around in a Python interactive interpreter to rework the existing XML template for CSLC into the configuration YAML - or at least the starting point to one. It's kinda hacky, and may need some tweaking to work on other PGEs, but I thought I'd share it here so you can maybe avoid the tedious task of translating the product spec documents + templates into configuration files by hand.

```python
>>> from bs4 import BeautifulSoup
>>> from yaml import dump, CDumper
>>>
>>> soup = BeautifulSoup(open('OPERA_ISO_metadata_L2_CSLC_S1_template.xml.jinja2'), 'xml')
>>> aa_s = soup.find('eos:AdditionalAttributes')
[...]

PR #537, opened 2024-10-17
Rollout Jinja2 template improvements to DISP-S1 PGE

RKuttruff:
## Description
- Rolled out templating improvements done in #496 to DISP-S1 PGE

## Affected Issues
- Completes #524 
  - Part of #520

## Testing
- Manually diffed generated XML files from unit tests
- Ran Int test in forward mode on `opera-dev-pge` and reviewed XML output file


src/opera/pge/disp_s1/disp_s1_pge.py
```diff
@@ -27,7 +28,10 @@
 from opera.util.h5_utils import get_disp_s1_product_metadata
 from opera.util.input_validation import validate_algorithm_parameters_config, validate_disp_inputs
 from opera.util.render_jinja2 import render_jinja2
-from opera.util.time import get_time_for_filename
+from opera.util.time import get_time_for_filename, get_catalog_metadata_datetime_str
+
+
+MEASURED_PARAMETER_PATH_SEPARATOR = '/'
```
collinss-jpl, line 34: Is this constant defined in multiple places at this point? Maybe we should refactor to a single location in `h5_util.py`, since it's only needed when defining the measured parameter config for PGE's that produce HDF5 outputs.

src/opera/pge/disp_s1/disp_s1_pge.py
```diff
@@ -589,6 +593,33 @@ def _qa_log_filename(self):
         """
         return self._ancillary_filename() + ".qa.log"
 
+    def augment_measured_parameters(self, measured_parameters):
```
collinss-jpl, line 596: Please add a docstring for this method

src/opera/pge/disp_s1/disp_s1_pge.py
```diff
@@ -589,6 +593,33 @@ def _qa_log_filename(self):
...
+        descriptions_file = self.runconfig.iso_measured_parameter_descriptions
+
+        new_measured_parameters = {}
+
+        if descriptions_file is not None:
```
collinss-jpl, line 601: Wondering if it makes more sense to just test `if descriptions_file`, so if somehow we're passed an empty string we still trigger the error case.

src/opera/pge/disp_s1/disp_s1_pge.py
```diff
@@ -589,6 +593,33 @@ def _qa_log_filename(self):
...
+                   'DISP-S1 metadata')
+            self.logger.critical(self.name, ErrorCode.ISO_METADATA_DESCRIPTIONS_CONFIG_NOT_FOUND, msg)
+
+        for parameter_var_name in descriptions:
+            key_path = parameter_var_name.split(MEASURED_PARAMETER_PATH_SEPARATOR)
+
+            mp = measured_parameters
+
+            while len(key_path) > 0:
+                mp = mp[key_path.pop(0)]
```
collinss-jpl, line 615: This could use a try/except block that catches `KeyError` and logs a descriptive error containing the full `parameter_var_name`. Will be helpful to catch typos in the description file down the line.

src/opera/pge/disp_s1/disp_s1_pge.py
```diff
@@ -609,6 +640,22 @@ def _collect_disp_s1_product_metadata(self, disp_product):
...
+                'ProductLevel': 3,
+                'ProductType': 'DISP-S1',
+                'ProductSource': 'Sentinel-1',
+                'ProcessingDateTime': get_catalog_metadata_datetime_str(self.production_datetime)
+            }
+
+            output_product_metadata['MeasuredParameters'] = self.augment_measured_parameters(output_product_metadata)
+
+            # Remove the hardcoded dict from main output product metadata dict once it's served its purpose
+            del output_product_metadata['static']
```
collinss-jpl, line 658: Is this really necessary? Does leaving it in cause unintended side-effects? If so, would be good to know that in the comment here.

collinss-jpl (changes requested):

src/opera/pge/disp_s1/disp_s1_pge.py
RKuttruff: Done

src/opera/pge/disp_s1/disp_s1_pge.py
RKuttruff: Done

src/opera/pge/disp_s1/disp_s1_pge.py
RKuttruff: Did this here and a couple of these other suggestions to CSLC as well as the code is identical there

src/opera/pge/disp_s1/disp_s1_pge.py
RKuttruff: I didn't really think it would be, I was thinking to minimize changes to that dictionary just in case there would be side-effects. I removed it for now

RKuttruff: Ran a diff and there's no issues with this statement gone

collinss-jpl (approved):

src/opera/util/h5_utils.py
```diff
@@ -32,6 +32,8 @@
     osr = MockOsr                           # pragma: no cover
 # pylint: enable=import-error,invalid-name
 
+MEASURED_PARAMETER_PATH_SEPARATOR = '/'
```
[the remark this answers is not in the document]
RKuttruff: I was going to do that but I couldn't figure out how to word it...

How does this sound: 
```
Character used to delimit HDF5 metadata subgroup "paths" in measured parameter config YAML files
```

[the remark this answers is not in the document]
collinss-jpl: Sounds great!

PR #541, opened 2024-10-30
RTC-S1 ISO XML Measured Parameters Rollout

### LINKED ISSUE #522: ISO XML Measured Parameters Update for RTC-S1
- [ ] Creation of the Measured Parameters Description YAML config 
  - Latest product spec can be found in Artifactory under the following path: `general/gov/nasa/jpl/opera/adt/rtc_s1/r5.1/final/documents/OPERA_RTC_S1_R5.1_Product_Specs.docx`
- [ ] Update the existing ISO XML template to use a [jinja2 for-loop construct](https://github.com/nasa/opera-sds-pge/blob/main/src/opera/pge/dswx_s1/templates/OPERA_ISO_metadata_L3_DSWx_S1_template.xml.jinja2#L586-L608) for the MeasuredParameters section. This should be performed after creation of the Parameter Description file described above

- [ ] Update `_collect_rtc_s1_product_metadata()` to include a call to augment_measured_parameters() to ensure all [metadata is formatted as expected](https://github.com/nasa/opera-sds-pge/blob/b5cb72869d440247e0e5e9a3fec0d25840abefdc/src/opera/pge/dswx_s1/dswx_s1_pge.py#L448) before the template is instantiated

- [ ] After all changes are made, run the integration test for the PGE (either via Jenkins or locally on dev machine) and inspect the resulting ISO XML product to ensure `MeasuredParameters` section is filled out as expected.

collinss-jpl:
## Description
- This branch rolls out the Measured Parameters ISO XML updates to the RTC-S1 PGE

## Affected Issues
- Resolves #522 

## Testing
- Branch has been tested to ensure generated ISO XML lines up with version generated by last release (v2.1.1) for both standard and static layer products


src/opera/pge/rtc_s1/rtc_s1_pge.py
```diff
@@ -695,6 +698,59 @@ def _qa_log_filename(self):
...
+            mp = measured_parameters
+
+            while len(key_path) > 0:
+                try:
+                    mp = mp[key_path.pop(0)]
+                    new_measured_parameters[parameter_var_name] = mp
+                except KeyError as err:
+                    msg = (f'Measured parameters contains no entry for description '
+                           f'key "{err}"')
+                    self.logger.warning(self.name, ErrorCode.ISO_METADATA_NO_ENTRY_FOR_DESCRIPTION, msg)
```
RKuttruff, line 748: Would it make sense to switch this error in the other PGEs that override the augment function to use this error code & log level?

src/opera/pge/base/base_pge.py
```diff
@@ -681,9 +684,15 @@ def augment_measured_parameters(self, measured_parameters):
             data_type = descriptions[name].get('attribute_data_type', guessed_data_type)
             attr_type = descriptions[name].get('attribute_type', "!Not Found!")
             attr_name = descriptions[name].get('display_name', guessed_attr_name)
+            escape_html = descriptions[name].get('escape_html', False)
 
-            augmented_parameters[name] = (dict(name=attr_name, value=value, attr_type=attr_type,
-                                               attr_description=attr_description, data_type=data_type))
+            if escape_html:
+                value = html.escape(value)
```
RKuttruff, line 690: If I understand correctly, this is to ensure that the generated XML isn't broken by special characters &c in the metadata. If that's the case, should we just escape/sanitize the value by default?

RKuttruff (approved):
Everything looks good. Just have a couple questions.

src/opera/pge/rtc_s1/rtc_s1_pge.py
collinss-jpl: It made sense for this PGE since the "standard" measured parameters are a superset of the parameters for "static layer" products, so a single description config could be used for both workflows. But to support that it was important that detection of a missing parameter did not cause a hard crash.

Since this is not necessarily the case for other PGEs like CSLC-S1, I'm not sure its appropriate to use the same logic there.

src/opera/pge/base/base_pge.py
collinss-jpl: I would prefer to take an explicit approach with escaping for only those fields that we're certain will contain a URL. I worry about potential unintended side-effects by applying it to every parameter.

src/opera/pge/rtc_s1/rtc_s1_pge.py
RKuttruff: Ah ok that makes sense!

Issue #556, opened 2024-12-11
[Bug]: Jenkins release pipeline failing due to running out of disk space


RKuttruff:
### Checked for duplicates

Yes - I've already checked

### Describe the bug

When running the release pipeline for DISP-S1 3.0.1, image tarball saving for upload failed due to running out of disk space.

It seems a lot of jenkins work is being stored outside of our `/data` volume.

I was able to work around it by replaying the pipeline with the path of the image tarball edited to be in `/data/tmp/`

### What did you expect?

I expected the pipeline to run as normal with no failures related to disk space
To the extent possible, update all existing Jenkinsfiles to store files in `/data/tmp/jenkins` instead of `/var/lib/jenkins`.
The delete-temp-dirs pipeline should also be updated to remove `/data/tmp/jenkins` as part of its periodic cleanup.

### Reproducible steps

```bash
Run the pipeline normally
```

### Environment

```bash
opera-pge-ci
```

RKuttruff:
opera-pge-ci linked jenkins dir to new dir under `/data`
Also updating jenkins pipelines to be more aggressive in conda cleaning

PR #558, opened 2024-12-19
Update Jenkins pipelines to be more aggressive in cleaning up build temp data

RKuttruff:
## Description
- Clean up conda package cache when environments are uninstalled
- Ensure the conda environment removal post steps are always run, even if prior docker cleanup script step fails
- Stop the cleanup pipeline from failing if conda install dir doesn't exist (doesn't have much effect re disk usage but the pipeline shouldn't fail if it has nothing to clean)

## Affected Issues
- #556 

## Testing
- Ran int test pipelines in sandbox

### DIFF
--- a/.ci/jenkins/build-int-test/Jenkinsfile
+++ b/.ci/jenkins/build-int-test/Jenkinsfile
@@ -64,10 +64,10 @@ pipeline {
             steps{
                 sh label: 'Download and install',
                    script: """#!/bin/bash
-                     mkdir -p /var/lib/jenkins/conda_installs/${RUN_ID}/miniconda3
-                     wget -q https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-x86_64.sh -O /var/lib/jenkins/conda_installs/${RUN_ID}/miniconda3/miniconda.sh
-                     bash /var/lib/jenkins/conda_installs/${RUN_ID}/miniconda3/miniconda.sh -b -u -p /var/lib/jenkins/conda_installs/${RUN_ID}/miniconda3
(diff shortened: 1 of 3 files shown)


.ci/jenkins/build-int-test/Jenkinsfile
```diff
@@ -142,13 +142,20 @@ pipeline {
...
         always {
-            echo "Cleaning up Docker images from local host"
-            sh ".ci/scripts/util/cleanup.sh ${DOCKER_TAG}"
-            echo "Removing int test conda environment"
-            sh "/var/lib/jenkins/conda_installs/${RUN_ID}/miniconda3/bin/conda env remove --yes --name int_test_env"
-            sh "/var/lib/jenkins/conda_installs/${RUN_ID}/miniconda3/bin/conda init --reverse"
-            sh "rm -rf /var/lib/jenkins/conda_installs/${RUN_ID}/"
-            deleteDir()
+            script {
+                catchError (buildResult: 'FAILURE', stageResult: 'FAILURE') {
```
collinss-jpl, line 146: Nice find, was just looking at the documentation for `catchError` and it looks like we could also include a message:

> message : String (optional)
> A message that will be logged to the console if an error is caught. If the stage result is specified, the message will also be associated with that result and may be shown in visualizations.

Might be worth adding to make it more obvious when either of the cleanup routines fails.

collinss-jpl (approved):

.ci/jenkins/build-int-test/Jenkinsfile
RKuttruff: Done

Issue #559, opened 2025-01-08
Input/Output Product Validation for DIST-S1


collinss-jpl:
The goal of this ticket is to add implementations to the DIST-S1 pre- and post-processor mixins that validate the existence and validity of the input and output products to DIST-S1.

In the pre-processor mixin class, we need to implement an input validation method that examines the following portions of the DIST-S1 RunConfig:

- `pre_rtc_copol`
- `pre_rtc_crosspol`
- `post_rtc_copol`
- `post_rtc_crosspol`

For each of the input file paths listed in the above sections, we want to validate the following:

1. File location exists
2. File is non-zero in size
3. File has an expected extension (.tif)

 The `util.input_validation.check_input` function should be reused to accomplish these checks.

In the post-processor mixin class, we need a similar validation method, but for the output products. Note that for the current SAS interface delivery, output products are always written to the current working directory within the container, so the output validation method may make the same assumption for now.

Based on the current interface, we can expect the SAS to create a single output directory ( ex: `OPERA_L3_DIST-ALERT-S1_T10SGD_20241103T015902Z_20241204T175000Z_S1_30_v0.0.1`) containing multiple output products for the single granule:

```
OPERA_L3_DIST-ALERT-S1_T10SGD_20241103T015902Z_20241204T175000Z_S1_30_v0.0.1_DATE-FIRST.tif       OPERA_L3_DIST-ALERT-S1_T10SGD_20241103T015902Z_20241204T175000Z_S1_30_v0.0.1_GEN-METRIC.tif
OPERA_L3_DIST-ALERT-S1_T10SGD_20241103T015902Z_20241204T175000Z_S1_30_v0.0.1_DATE-LATEST.tif      OPERA_L3_DIST-ALERT-S1_T10SGD_20241103T015902Z_20241204T175000Z_S1_30_v0.0.1_N-DIST.tif
OPERA_L3_DIST-ALERT-S1_T10SGD_20241103T015902Z_20241204T175000Z_S1_30_v0.0.1_DIST-STATUS-ACQ.tif  OPERA_L3_DIST-ALERT-S1_T10SGD_20241103T015902Z_20241204T175000Z_S1_30_v0.0.1_N-OBS.tif
OPERA_L3_DIST-ALERT-S1_T10SGD_20241103T015902Z_20241204T175000Z_S1_30_v0.0.1_DIST-STATUS.tif
```

[...]

PR #561, opened 2025-01-13
DIST-S1 Input/Output Validations

RKuttruff:
## Description
- File validations for PGE inputs and outputs

## Affected Issues
- #559 

## Testing
- Added unit tests for all checks in validations

### DIFF
--- a/src/opera/pge/dist_s1/dist_s1_pge.py
+++ b/src/opera/pge/dist_s1/dist_s1_pge.py
@@ -95,12 +95,12 @@ class DistS1PostProcessorMixin(PostProcessorMixin):
         product_id = re.compile(product_id_pattern + r'$')
         granule_filename = re.compile(granule_filename_pattern)
 
-        oputput_product_path = abspath(self.runconfig.output_product_path)
+        output_product_path = abspath(self.runconfig.output_product_path)
         output_products = []
 
-        for file in os.listdir(oputput_product_path):
-            dir_path = join(oputput_product_path, file)
-            if isdir(dir_path) and product_id.match(file) is not None:
+        for file in os.listdir(output_product_path):
+            dir_path = join(output_product_path, file)
+            if isdir(dir_path) and product_id.match(file):
                 bands = []
                 generated_band_names = []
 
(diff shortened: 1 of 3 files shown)


src/opera/pge/dist_s1/dist_s1_pge.py
```diff
@@ -55,19 +69,90 @@ class DistS1PostProcessorMixin(PostProcessorMixin):
...
+        granule_filename_pattern = (product_id_pattern + rf'((_(?P<layer_name>{"|".join(self._output_layer_names)}))|'
+                                                         r'_BROWSE)?[.](?P<ext>tif|tiff|png)$')
+
+        product_id = re.compile(product_id_pattern + r'$')
+        granule_filename = re.compile(granule_filename_pattern)
+
+        oputput_product_path = abspath(self.runconfig.output_product_path)
+        output_products = []
+
+        for file in os.listdir(oputput_product_path):
```
collinss-jpl, line 101: typo here: oputput_product_path -> output_product_path

src/opera/pge/dist_s1/dist_s1_pge.py
```diff
@@ -55,19 +69,90 @@ class DistS1PostProcessorMixin(PostProcessorMixin):
...
+            dir_path = join(oputput_product_path, file)
+            if isdir(dir_path) and product_id.match(file) is not None:
```
collinss-jpl, line 103: I think the `is not None` portion can be omitted and it will read a bit cleaner

src/opera/pge/dist_s1/dist_s1_pge.py
```diff
@@ -55,19 +69,90 @@ class DistS1PostProcessorMixin(PostProcessorMixin):
...
+
+                for granule in os.listdir(dir_path):
+                    granule_path = join(dir_path, granule)
+                    if isfile(granule_path):
+                        match_result = granule_filename.match(granule)
+
+                        if match_result is None:  # or match_result.groupdict()['ext'] != 'tif':
+                            error_msg = f'Invalid product filename {granule}'
+                            self.logger.critical(self.name, ErrorCode.INVALID_OUTPUT, error_msg)
+                        else:
```
collinss-jpl, line 115: I think we can omit this `else` branch and promote the conditional within to be part of the `if match_result is None` conditional, i.e.:

```python
if match_result is None:
   ...
elif os.stat(granule_path).st_size == 0:
   ...
elif match_result.groupdict()['layer_name'] not in self._output_layer_names:
   ...
else:
   bands.append(granule_path)
   ...
```

Should help to bring the complexity of this conditional down a bit. The final `else` can also be omitted since the `self.logger.critical` calls result in an exception.

src/opera/pge/dist_s1/dist_s1_pge.py
```diff
@@ -55,19 +69,90 @@ class DistS1PostProcessorMixin(PostProcessorMixin):
...
+                                error_msg = f'Invalid layer name "{match_result.groupdict()["layer_name"]}" in output.'
+                                self.logger.critical(self.name, ErrorCode.INVALID_OUTPUT, error_msg)
+                            else:
+                                bands.append(granule_path)
+                                generated_band_names.append(match_result.groupdict()['layer_name'])
+
+                if len(bands) != len(self._output_layer_names):
+                    error_msg = f'Incorrect number of output bands generated: {len(generated_band_names)}'
+                    self.logger.critical(self.name, ErrorCode.INVALID_OUTPUT, error_msg)
+                else:
```
collinss-jpl, line 129: I also think this `else` is unnecessary, since a `self.logger.critical` call will raise an exception

src/opera/pge/dist_s1/dist_s1_pge.py
```diff
@@ -55,19 +69,90 @@ class DistS1PostProcessorMixin(PostProcessorMixin):
...
+            self.logger.critical(self.name, ErrorCode.INVALID_OUTPUT, error_msg)
+
+        #  "(?P<id>(?P<project>OPERA)_(?P<level>L3)_(?P<product_type>DIST)-(?P<source>S1)_(?P<tile_id>T[^\\W_]{5})_(?P<acquisition_ts>(?P<acq_year>\\d{4})(?P<acq_month>\\d{2})(?P<acq_day>\\d{2})T(?P<acq_hour>\\d{2})(?P<acq_minute>\\d{2})(?P<acq_second>\\d{2})Z)_(?P<creation_ts>(?P<cre_year>\\d{4})(?P<cre_month>\\d{2})(?P<cre_day>\\d{2})T(?P<cre_hour>\\d{2})(?P<cre_minute>\\d{2})(?P<cre_second>\\d{2})Z)_(?P<sensor>S1[AC])_(?P<spacing>30)_(?P<product_version>v\\d+[.]\\d+[.]\\d+))$",
+        #  (?P<id>(?P<project>OPERA)_(?P<level>L3)_(?P<product_type>DIST)-(?P<source>S1)_(?P<tile_id>T[^\W_]{5})_(?P<acquisition_ts>(?P<acq_year>\d{4})(?P<acq_month>\d{2})(?P<acq_day>\d{2})T(?P<acq_hour>\d{2})(?P<acq_minute>\d{2})(?P<acq_second>\d{2})Z)_(?P<creation_ts>(?P<cre_year>\d{4})(?P<cre_month>\d{2})(?P<cre_day>\d{2})T(?P<cre_hour>\d{2})(?P<cre_minute>\d{2})(?P<cre_second>\d{2})Z)_(?P<sensor>S1[AC])_(?P<spacing>30)_(?P<product_version>v\d+[.]\d+[.]\d+))((_(?P<layer_name>DIST-GEN-STATUS|DIST-GEN-STATUS-ACQ|GEN-METRIC|DATE-FIRST|DATE-LATEST|N-DIST|N-OBS))|_BROWSE)?[.](?P<ext>tif|tiff|png|iso\.xml)$
```
collinss-jpl, line 137: Are these for reference? Can they be removed?

src/opera/pge/dist_s1/dist_s1_pge.py
```diff
@@ -55,19 +69,90 @@ class DistS1PostProcessorMixin(PostProcessorMixin):
...
+                                generated_band_names.append(match_result.groupdict()['layer_name'])
+
+                if len(bands) != len(self._output_layer_names):
+                    error_msg = f'Incorrect number of output bands generated: {len(generated_band_names)}'
+                    self.logger.critical(self.name, ErrorCode.INVALID_OUTPUT, error_msg)
+                else:
+                    output_products.append(dir_path)
+
+        if len(output_products) != 1:
+            error_msg = f'Incorrect number of output products generated: {len(output_products)}'
```
collinss-jpl, line 133: For clarity, let's use the term "output granules" here since we're talking about directories rather than products (the individual files _within_ a granule).

src/opera/test/pge/dist_s1/test_dist_s1_pge.py
```diff
@@ -108,16 +166,255 @@ def test_dist_s1_pge_execution(self):
         expected_log_file = pge.logger.get_file_name()
         self.assertTrue(os.path.exists(expected_log_file))
 
-        # Lastly, check that the dummy output products were created
-        tif_files = glob.glob(join(pge.runconfig.output_product_path, "*.tif"))
-        self.assertEqual(len(tif_files), 1)
```
collinss-jpl, line 113: Why was this removed and not just updated to test for the actual number of output tifs we expect?

src/opera/test/pge/dist_s1/test_dist_s1_pge.py
```diff
@@ -74,6 +130,8 @@ def test_dist_s1_pge_execution(self):
         """
         runconfig_path = join(self.data_dir, 'test_dist_s1_config.yaml')
 
+        DistS1PgeTestCase._create_dummy_input_files(runconfig_path)
```
collinss-jpl, line 133: I think we could make this call part of `setUp` so we don't have to remember to call it in every test case that needs it (which is likely all of them anyhow)

collinss-jpl (changes requested):
Logic all looks good to me, but things can be tided up a bit to reduce complexity of some of the conditional statements.

src/opera/pge/dist_s1/dist_s1_pge.py
RKuttruff: Is that for the error message or should that be reflected in variable names too?

src/opera/test/pge/dist_s1/test_dist_s1_pge.py
RKuttruff: I have no idea how those lines got deleted....
Definitely didn't intend to alter that case or don't recall if/why I would

RKuttruff:
@collinss-jpl Made requested changes

src/opera/pge/dist_s1/dist_s1_pge.py
```diff
@@ -55,19 +69,86 @@ class DistS1PostProcessorMixin(PostProcessorMixin):
...
+        validates the output product files and performs other base
+        postprocessing steps (writing log file, etc).
 
         Parameters
         ----------
         **kwargs: dict
             Any keyword arguments needed by the post-processor
         """
-        super().run_postprocessor(**kwargs)
+        # super().run_postprocessor(**kwargs)
```
collinss-jpl, line 147: Whoops, missed this but it should be removed

collinss-jpl (approved):
Found one commented out line that should be removed, but other wise this looks great! Feel free to merge after removing said line and the branch makes it through CI.

RKuttruff:
Will do & thanks!

Issue #571, opened 2025-01-29
[Bug]: DISP-S1 ISO XML Additional issues


RKuttruff:
### Checked for duplicates

Yes - I've already checked

### Describe the bug

Several issues were observed:
- Some measured parameters' values need to be escaped (Invalid XML)
- Optional parameters found to be missing are still shoved into the output XML with an incorrect value of a JSON-rendered dict of the parent group (This bug will also apply to CSLC-S1 and ~~RTC-S1~~ (ETA: Fix is already present in RTC))
- References to troposphere files in the DISP-S1 XML template

### What did you expect?

ISO XML to be valid with sensible attribute values and nothing missing at render-time

### Reproducible steps

```bash
1. Run latest release of DISP-S1 PGE
```

### Environment

```bash

```

PR #573, opened 2025-01-29
DISP-S1 ISO XML patches

RKuttruff:
## Description
- Escape some measured parameter values
- Fix bug with missing MPs not being skipped properly
  - Also fixed in CSLC-S1; fix was already present in RTC-S1
- Removed obsolete reference to troposphere files from template

## Affected Issues
- #571 

## Testing
- Ran integration test on dev-pge and inspected ISO XML file and PGE log file

### DIFF
--- a/src/opera/pge/cslc_s1/cslc_s1_pge.py
+++ b/src/opera/pge/cslc_s1/cslc_s1_pge.py
@@ -646,6 +646,7 @@ class CslcS1PostProcessorMixin(PostProcessorMixin):
             while len(key_path) > 0:
                 try:
                     mp = mp[key_path.pop(0)]
+                    new_measured_parameters[parameter_var_name] = mp
                 except KeyError as e:
                     msg = (f'Measured parameters configuration contains a path {parameter_var_name} that is missing '
                            f'from the output product')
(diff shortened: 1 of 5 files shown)


RKuttruff:
Found that my approach for handling missing optionals still didn't work - rerunning int test

collinss-jpl (commented):
The approach looks fine, but seeing how you had to update three nearly identical versions of `augment_measured_parameters` (the version that's specialized for working with HD5 metadata paths), I'm wondering if we should take this opportunity to promote this HDF5 version to `base_pge.py`. The overloaded versions of `augment_measured_parameters` in each PGE module would then just invoke the shared implementation in `base_pge`. What do you think @RKuttruff?

RKuttruff:
@collinss-jpl As in replace the overloaded `augment_measured_parameters` with a new, common method in `base_pge`?

collinss-jpl:
> @collinss-jpl As in replace the overloaded `augment_measured_parameters` with a new, common method in `base_pge`?

After thinking it through a bit more, lets table this discussion for now. There's definitely refactoring potential here to remove some duplicated code, but I think it should be handled in a separate ticket.

collinss-jpl (approved):

Issue #572, opened 2025-01-29
[New Feature]: Jinja render optional validation step


RKuttruff:
### Checked for duplicates

Yes - I've already checked

### Alternatives considered

Yes - and alternatives don't suffice

### Related problems

Should pick up on XML issues (#571) during integration testing and not further down the line with manual inspection.

### Describe the feature request

Add an optional "validator" parameter to `opera.util.render_jinja2.render_jinja2` that is a callable that takes the rendered text and tries to parse it, raising an error if it cannot be parsed.

Examples:
```python
from opera.util.render_jinja2 import render_jinja2

import json
import yaml
from lxml import etree
from functools import partial

# Render JSON
render_jinja2(JSON_TEMPLATE, JSON_DATA, validator=json.loads)

# Render YAML
render_jinja2(YAML_TEMPLATE, YAML_DATA, validator=partial(yaml.load, Loader=yaml.SafeLoader))

# Render XML
render_jinja2(XML_TEMPLATE, XML_DATA, validator=lambda x: etree.fromstring(x.encode('utf-8')))
```

collinss-jpl:
Resolved by #575

PR #575, opened 2025-01-30
Optional parameter to validate parsability of rendered text ( + RTC-S1 MPC patch)

RKuttruff:
## Description
- Add optional parameter to `opera.util.render_jinja2.render_jinja2` that, if provided, will be used to attempt to parse the rendered output, raising an exception if the output is invalid. The parameter is a callable that takes a single positional argument of the rendered text string and is expected to raise an exception if the text is invalid. Eg.: `json.loads`, `yaml.load`
- Added XML validator to PGEs that produce ISO XML documents

## Affected Issues
- #572 
- #577 

## Testing
- Added unit tests for both valid and invalid rendered JSON, YAML, and XML
- Int tests on PGEs with implemented validations:
  - [x] CSLC-S1
  - [x] DISP-S1
  - [x] DSWx-HLS
  - [x] DSWx-S1
  - [x] DSWx-NI
  - [x] RTC-S1
    - ~~Baseline worked fine, missing parameter in static layers - will need addressing if new PGE version is released (#577)~~


src/opera/util/render_jinja2.py
```diff
@@ -109,7 +114,21 @@ def __getattr__(self, name):
...
+JSON_VALIDATOR = json.loads
+YAML_VALIDATOR = partial(yaml.load, Loader=yaml.SafeLoader)
+XML_VALIDATOR = _try_parse_xml_string
+
+
+def render_jinja2(
+        template_filename: str,
+        input_data: dict,
+        logger: PgeLogger = None,
+        validator=None
```
collinss-jpl, line 130: I think I would rather see this parameter default to `XML_VALIDATOR` and have a caller decide to pass `None` if they want to disable validation. Then we won't need the changes to each individual PGE module in this branch.

src/opera/util/render_jinja2.py
```diff
@@ -155,10 +177,17 @@ def render_jinja2(template_filename: str, input_data: dict, logger: PgeLogger =
 
     rendered_text = template.render(input_data)
 
+    if validator is not None:
+        try:
+            validator(rendered_text)
+        except Exception as err:
+            raise RuntimeError(f"Error rendering Jinja2 template: {err}") from err
```
collinss-jpl, line 184: Lets change this to "Error parsing rendered ISO XML" so its a little more obvious that the failure is during parsing, not creation of the XML from the template.

collinss-jpl (commented):
Had a few comments, but since testing of the branch uncovered #577, can we just include the fix of marking the problematic fields as optional within this branch?

RKuttruff:
@collinss-jpl Since the static workflow will _always_ omit those parameters, I think it might be better to make a static version of the MPC like we did for CSLC-S1

collinss-jpl (approved):

PR #581, opened 2025-02-04
Refactor Measured Parameters Augmentation Functions

### LINKED ISSUE #579: Refactor Measured Parameter methods to render_jinja2.py
The new approach to using the measured parameters config to render the ISO XML products has been working well, however, its resulted in a bit of code duplication between the PGE's that extract metadata from HDF5/NetCDF products.

To remedy this, I propose that the current implementations of `augment_measured_parameters`, both in `base_pge.py` as well as the overloaded functions in CSLC/RTC etc., be refactored into utility functions within the `render_jinja2.py` module.

To accomplish this:

- [x] Refactor `base_pge.augment_measured_parameters()` to `render_jinja2.py` (no name change needed)
  - The path to the descriptions file and an instance of `PgeLogger` should be come arguments to the refactored version
- [x] Refactor `cslc_s1_pge.augment_measured_parameters()` (or any single version) to `render_jinja2.py`, but rename to `augment_hd5_measured_parameters()`.
  - The signature should be updated the same as `augment_measured_parameters()` to take the path to the description file and a logger
  - This version should still invoke the baseline `augment_measured_parameters()`
  - Update logged messages to remove any PGE-specific references (such as PGE name)
- [x] Update PGE modules to reference the appropriate function when collecting metadata to render the ISO XML with

RKuttruff:
## Description
- Move the base implementation of `augment_measured_parameters` and its specialized overrides in CSLC-S1, RTC-S1 and DISP-S1 into the `opera.util.render_jinja2` module

## Affected Issues
- Resolves #579 

## Testing
- Verified unit tests still pass (local)
- Ran int tests on opera-dev-pge for PGEs using `augment_measured_parameters` (DSWx-HLS) and `augment_hd5_measured_parameters` (CSLC-S1) with images built off the main branch and this PR and diff'd an iso XML for a sample product from each to validate no significant changes/regressions in template filling occurred
- Some more PGE int tests to check nothing else broke (Particularly RTC-S1, DISP-S1 and DSWx-NI)

### DIFF
--- a/src/opera/pge/disp_s1/disp_s1_pge.py
+++ b/src/opera/pge/disp_s1/disp_s1_pge.py
@@ -26,7 +26,7 @@ from opera.util.error_codes import ErrorCode
 from opera.util.h5_utils import get_cslc_s1_product_metadata
 from opera.util.h5_utils import get_disp_s1_product_metadata
 from opera.util.input_validation import validate_algorithm_parameters_config, validate_disp_inputs
-from opera.util.render_jinja2 import augment_hd5_measured_parameters, render_jinja2
+from opera.util.render_jinja2 import augment_hdf5_measured_parameters, render_jinja2
 from opera.util.time import get_time_for_filename, get_catalog_metadata_datetime_str
 
 
@@ -621,7 +621,7 @@ class DispS1PostProcessorMixin(PostProcessorMixin):
                 'ProcessingDateTime': get_catalog_metadata_datetime_str(self.production_datetime)
             }
 
-            output_product_metadata['MeasuredParameters'] = augment_hd5_measured_parameters(
+            output_product_metadata['MeasuredParameters'] = augment_hdf5_measured_parameters(
                 output_product_metadata,
                 self.runconfig.iso_measured_parameter_descriptions,
                 self.logger
--- a/src/opera/pge/disp_s1/disp_s1_pge.py
+++ b/src/opera/pge/disp_s1/disp_s1_pge.py
@@ -28,7 +28,7 @@ from opera.util.h5_utils import get_cslc_s1_product_metadata
 from opera.util.h5_utils import get_disp_s1_product_metadata
 from opera.util.h5_utils import MEASURED_PARAMETER_PATH_SEPARATOR
 from opera.util.input_validation import validate_algorithm_parameters_config, validate_disp_inputs
-from opera.util.render_jinja2 import render_jinja2
+from opera.util.render_jinja2 import augment_hd5_measured_parameters, render_jinja2
 from opera.util.time import get_time_for_filename, get_catalog_metadata_datetime_str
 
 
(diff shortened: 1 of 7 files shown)


src/opera/pge/disp_s1/disp_s1_pge.py
```diff
@@ -18,17 +18,15 @@
...
 from opera.pge.base.base_pge import PostProcessorMixin
 from opera.pge.base.base_pge import PreProcessorMixin
 from opera.util.dataset_utils import parse_bounding_polygon_from_wkt
 from opera.util.error_codes import ErrorCode
 from opera.util.h5_utils import get_cslc_s1_product_metadata
 from opera.util.h5_utils import get_disp_s1_product_metadata
-from opera.util.h5_utils import MEASURED_PARAMETER_PATH_SEPARATOR
 from opera.util.input_validation import validate_algorithm_parameters_config, validate_disp_inputs
-from opera.util.render_jinja2 import render_jinja2
+from opera.util.render_jinja2 import augment_hd5_measured_parameters, render_jinja2
```
collinss-jpl, line 29: I just realized I made a typo in the ticket description, I had intended this function to be named `augment_hdf5_measured_parameters`, sorry about that!

RKuttruff: I flipped a coin as to whether that was a typo and it came up wrong I guess :)

src/opera/util/render_jinja2.py
```diff
@@ -202,6 +204,132 @@ def python_type_to_xml_type(obj) -> str:
...
+    file.
+
+    For HDF5 or NetCDF metadata, use augment_hd5_measured_parameters()
+
+    Parameters
+    ----------
+    measured_parameters : dict
+       The GeoTIFF metadata from the output product. See get_geotiff_metadata()
+    mpc_path: str
+        Path to the Measured Parameters Descriptions YAML file
```
collinss-jpl, line 220: Looking over how this function works again, and I think we need to document the following:

This function can accept `None` for `mpc_path`, in which case all descriptions will be filled in with `Not Provided`. Otherwise, any missing descriptions are filled in with `!Not Found!`, indicating a configuration error.

src/opera/util/render_jinja2.py
```diff
@@ -202,6 +204,132 @@ def python_type_to_xml_type(obj) -> str:
...
+    dictionary based on the "paths" provided in the variable keys of the configuration
+    YAML file. The result of this preprocessing is then safely passed to the base
+    method to get the correct structure expected by the Jinja template.
+
+    Parameters
+    ----------
+    measured_parameters : dict
+        The HDF5 metadata from the output product. See get_cslc_s1_product_metadata()
+    mpc_path: str
+        Path to the Measured Parameters Descriptions YAML file
```
collinss-jpl, line 290: I also think we should clarify that, unlike the baseline `augment_measured_parameters`, it is an error if `None` is provided or `mpc_path` does not exist.

collinss-jpl (approved):
Refactor looks great, just had some suggestions to improve docstrings (and sorry again about the miscommunication on the name for `augment_hdf5_measured_parameters`!)

src/opera/pge/disp_s1/disp_s1_pge.py
collinss-jpl: I wouldn't bet against my own stupidity if I were you! 😆 😢

RKuttruff:
Suggested changes have been made

PR #588, opened 2025-02-24
Fix bug in int test scripts for DSWx-HLS & DISP-S1

### LINKED ISSUE #587: [Bug]: DSWx-HLS Integration test script reports passing when first test case fails
### Checked for duplicates

Yes - I've already checked

### Describe the bug

When running the DSWx-HLS int test script, two executions are ran `l30_greenland` and `s30_louisiana` 

In the event that `l30_greenland` fails but `s30_louisiana` passes, the overall test reports a pass

I suspect it's due to this block: https://github.com/nasa/opera-sds-pge/blob/0cbe9d93c59711cbd70ab56e61f99b4985c22e3e/.ci/scripts/dswx_hls/test_int_dswx_hls.sh#L111-L117

overwriting the exit code from the previous test case iteration.

### What did you expect?

I expected the overall test to report failure if either test case fails.

### Reproducible steps

```bash
On opera-dev-pge, build a PROTEUS SAS image off of commit e22fc8a695c274d25f0281b019d3fa2b265a816fe22fc8a695c274d25f0281b019d3fa2b265a816f, build the DSWx-HLS PGE on top of that. Run the int test, then modify the int script to only run l30_greenland and run the int test again.
```

### Environment

```bash
opera-dev-pge
```

RKuttruff:
## Description
- Fixes bug in int test scripts with more than one PGE invocation & comparison step. Previously, the final test outcome overwrote all previous outcomes, meaning any test failures in prior steps would be ignored if the final step was successful. 

## Affected Issues
- Fixes #587 

## Testing
- Tested on dev with DSWx-HLS PGE image with which the bug was first observed

### DIFF
--- a/.ci/scripts/disp_s1/test_int_disp_s1.sh
+++ b/.ci/scripts/disp_s1/test_int_disp_s1.sh
@@ -129,7 +129,11 @@ do
         overall_status=1
     else
         # Retrieve the return code written to disk by the comparison script
-        overall_status=$(cat "$output_dir/compare_disp_s1_products.rc")
+        test_status=$(cat "$output_dir/compare_dswx_hls_products.rc")
+
+        if [ $test_status -ne 0 ]; then
+          overall_status=$test_status
+        fi
     fi
 done
 
--- a/.ci/scripts/dswx_hls/test_int_dswx_hls.sh
+++ b/.ci/scripts/dswx_hls/test_int_dswx_hls.sh
@@ -45,6 +45,12 @@ test_int_setup_test_data
 # Setup cleanup on exit
 trap test_int_trap_cleanup EXIT
 
+# overall_status values and their meaning
+# 0 - pass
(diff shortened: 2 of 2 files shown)


.ci/scripts/disp_s1/test_int_disp_s1.sh
```diff
@@ -129,7 +129,11 @@ do
         overall_status=1
     else
         # Retrieve the return code written to disk by the comparison script
-        overall_status=$(cat "$output_dir/compare_disp_s1_products.rc")
+        test_status=$(cat "$output_dir/compare_dswx_hls_products.rc")
+
+        if [ $test_status -ne 0 ]; then
```
collinss-jpl, line 134: Just out of curiosity, can we accomplish this with a logical assignment operator? Something like `overall_status = $overall_status | $test_status`?

collinss-jpl (approved):

PR #592, opened 2025-02-26
586 TROPO stub implementation 

kevinmarlis:
## Description
- Implemented barebones versions of pre/post mixin classes and executor class inheriting from `base_pge`
- Added runconfig schema
- Added `TROPO_PGE` to `TropoExecutor` in `pge_main`
- Added `test_tropo_pge` and corresponding `test_tropo_config`
- Added dockerfile, build/test scripts, updated build_all/test_all scripts

## Affected Issues
- #549 

## Testing
- Added unit test module for TROPO PGE, successfully runs and validates runconfig schema
- Added CI script for unit testing


.ci/docker/Dockerfile_tropo
```diff
@@ -0,0 +1,67 @@
+# Dockerfile to produce the production TROPO PGE Docker image for OPERA
+# Authors: Kevin Marlis, Scott Collins, Ray Bambery
```
collinss-jpl, line 2: We don't need to include Ray's name here since this is technically a "new" file

src/opera/pge/tropo/tropo_pge.py
```diff
@@ -0,0 +1,90 @@
...
+    Main class for execution of the TROPO PGE, including the SAS layer.
+    This class essentially rolls up the TROPO-specific pre- and post-processor
+    functionality, while inheriting all other functionality for setup and execution
+    of the SAS from the base PgeExecutor class.
+    """
+
+    NAME = "TROPO"
+    """Short name for the TROPO PGE"""
+
+    LEVEL = "L4"
```
collinss-jpl, line 81: Is this right? Is L4 the level used in the documentation?

kevinmarlis: Yes, it's an L4 product according to the documentation:

> OPERA (Observational Products for End-users from Remote-sensing Analysis) Level-4 Troposphere RADAR Delays

collinss-jpl: Ah wasn't aware of that!

src/opera/pge/tropo/tropo_pge.py
```diff
@@ -0,0 +1,90 @@
...
+"""
+
+from opera.pge.base.base_pge import PgeExecutor
+from opera.pge.base.base_pge import PostProcessorMixin
+from opera.pge.base.base_pge import PreProcessorMixin
+
+
+class TROPOPreProcessorMixin(PreProcessorMixin):
+    """
+    Mixin class responsible for handling all pre-processing steps for the DSWX-NI
```
RKuttruff, line 17: Bit nitpicky, but there's a small cut/paste should be updated here

kevinmarlis: Whoops, thought I caught all of them. Thanks, Riley!

.ci/scripts/tropo/build_tropo.sh
```diff
@@ -0,0 +1,83 @@
...
+'
+
+PGE_NAME="tropo"
+IMAGE="opera_pge/${PGE_NAME}"
+BUILD_DATE_TIME=$(date -u +'%Y-%m-%dT%H:%M:%SZ')
+
+# defaults, SAS image should be updated as necessary for new image releases from ADT
+[ -z "${WORKSPACE}" ] && WORKSPACE=$(realpath $(dirname $(realpath $0))/../../..)
+[ -z "${TAG}" ] && TAG="${USER}-dev"
+[ -z "${SAS_IMAGE}" ] && SAS_IMAGE="artifactory-fn.jpl.nasa.gov:16001/gov/nasa/jpl/opera/adt/tropo:interface_0.1.tar"
```
collinss-jpl, line 27: A little bit of background: When ADT delivers SAS images to us in Artifactory, they actually delivery 2 different "form-factors" of the same image. One is the .tar image, which we usually use for Acceptance Testing since its a bit easier to download and load into the local Docker registry on the dev machine. However, they also deliver the image in a "raw" format, which is the version we use to define the base image for our PGE build scripts.

This is all to say that the correct default value here should be `SAS_IMAGE="artifactory-fn.jpl.nasa.gov:16001/gov/nasa/jpl/opera/adt/opera/tropo:interface_0.1"`

collinss-jpl (commented):
@kevinmarlis I've finished working with this branch on the dev machine, and I found I had to make the following minor tweaks to get everything to build and test correctly. I think once you've applied these to this branch it should be in good shape to merge. Thanks!

[tropo_stub_diff.txt](https://github.com/user-attachments/files/18999144/tropo_stub_diff.txt)

.ci/scripts/tropo/build_tropo.sh
kevinmarlis: That makes sense! Thanks for clarifying.

collinss-jpl (approved):
Everything looks good now, made a small update to include the new build/test scrips in our basic CI pipeline and everything passes. Great work @kevinmarlis!

PR #601, opened 2025-03-05
Fix for stale references in DISP-S1 ISO XML

### LINKED ISSUE #600: [Bug]: Stale references to "identification" dictionary in DIST-S1 ISO template
### Checked for duplicates

Yes - I've already checked

### Describe the bug

While examining the results of the DIST-S1 integration test, I noticed several of the following in the main PGE log:

`2025-03-04T21:21:29.682724Z, Warning, DIST_S1_PGE::base_pge.py, render_jinja2, 702002, /home/ops/opera/util/render_jinja2.py:93, "Missing/undefined ISO metadata template variable: 'dict object' has no attribute 'identification'"`

Looks like there are two leftover references to the "identification" dictionary we extract from some HDF5 products:
https://github.com/nasa/opera-sds-pge/blob/main/src/opera/pge/dist_s1/templates/OPERA_ISO_metadata_L3_DIST_S1_template.xml.jinja2#L533-L534

### What did you expect?

We will need different placeholders for the start/end interval of the associated DIST-S1 product, and we will also need to update the `_collect_dist_s1_product_metadata` method to ensure we assign values for the new placeholders.

If there is not a sensing start/end time currently within the product metadata, then for the time being we should use the acquisition time as parsed from the product filename for both the start and end time. The time(s) should be converted to ISO format (`2004-04-12T13:20:00Z`), see `util.time.get_iso_time()`

### Reproducible steps

```bash
Should be reproducible for any run of the current DIST-S1 PGE. Examine the log file for the warning messages described above.
```

### Environment

```bash
...
```

RKuttruff:
## Description
- Adds secondary zero doppler start/end times to top-level product metadata subdictionary and refers to that in ISO XML template

## Affected Issues
- Fixes #600 

## Testing
- Verified sections populated during local unit test execution
- ~~Verification with int test on dev-pge underway~~
- Passed int test on dev-pge (`/data/tmp/tmp.XjaXM3K6YO/output_disp_s1/historical`)

### DIFF
--- a/src/opera/pge/disp_s1/disp_s1_pge.py
+++ b/src/opera/pge/disp_s1/disp_s1_pge.py
@@ -618,7 +618,9 @@ class DispS1PostProcessorMixin(PostProcessorMixin):
                 'ProductLevel': 3,
                 'ProductType': 'DISP-S1',
                 'ProductSource': 'Sentinel-1',
-                'ProcessingDateTime': get_catalog_metadata_datetime_str(self.production_datetime)
+                'ProcessingDateTime': get_catalog_metadata_datetime_str(self.production_datetime),
+                'StartTime': output_product_metadata['identification']['secondary_zero_doppler_start_time'],
+                'EndTime': output_product_metadata['identification']['secondary_zero_doppler_end_time']
             }
 
             output_product_metadata['MeasuredParameters'] = augment_hdf5_measured_parameters(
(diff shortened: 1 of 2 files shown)


collinss-jpl (changes requested):
I like your approach @RKuttruff, unfortunately the issue is in the **DIST-S1** template, not DISP-S1 (easy mistake I realize)

RKuttruff:
🤦

RKuttruff:
Closing for a new branch

Issue #595, opened 2025-03-04
DIST-S1 Integration Test Framework Addition


collinss-jpl:
Ticket to implement the PGE/SAS Integration Test framework for DIST-S1. This should include:

- [x] Creation of the `compare_dist_s1_products.sh` (or `.py` if it makes more sense) that is used as the "Quality Assurance" application executed directly after the SAS during the integration test. This script should leverage the DIST-S1 product comparison feature described [here](https://github.com/opera-adt/dist-s1?tab=readme-ov-file#checking-if-dist-s1-products-are-equal)
- [X] Creation of the initial expected test input and output assets to be downloaded and used with the integration test. 
- [x] Creation of the RunConfig to be used with the test
  - Ensure the comparison script described above is configured within the `QAExecutable` section of this RunConfig 
- [x] Creation of the test_int_dist_s1.sh script which orchestrates download of the expected assets, metrics colletion, and execution of the DIST-S1 container
- [x] Update the `.ci/jenkins/build-test-deploy/Jenkinsfile` to include `dist_s1`

RKuttruff:
When testing out the comparison scripts, I used DSWx-S1 as a reference and when testing a failure in comparison, I noticed the script still reported PASSED

Then I noticed the QA script return code wasn't propagated to the int script in DSWx-S1 (it forces to 0 to avoid PGE failure)

DIST-S1: https://github.com/nasa/opera-sds-pge/blob/4721ea99e382d4dc41a26e058c8850f606354064/.ci/scripts/dist_s1/test_int_dist_s1.sh#L113-L123

DSWx-S1: https://github.com/nasa/opera-sds-pge/blob/4721ea99e382d4dc41a26e058c8850f606354064/.ci/scripts/dswx_s1/test_int_dswx_s1.sh#L115-L118

PR #604, opened 2025-03-06
Fix for stale references in DIST-S1 ISO XML

### LINKED ISSUE #600: [Bug]: Stale references to "identification" dictionary in DIST-S1 ISO template

RKuttruff:
## Description
- ACTUALLY addresses #600 (My bad!)
- No metadata field yet for acquisition time so I used the regexes in the output validation to parse the filenames
- Also included are some fixes to [an issue I found](https://github.com/nasa/opera-sds-pge/issues/595#issuecomment-2698718908) in the DSWx-S1 int test script and also found in the scripts for CSLC-S1 and DSWx-NI

## Affected Issues
- Fixes #600 

## Testing
- Unit tests passing both locally & on dev-pge
- Int tests for changed PGEs running on Jenkins (sandbox job 404 - passed)

### DIFF
--- a/src/opera/pge/dist_s1/dist_s1_pge.py
+++ b/src/opera/pge/dist_s1/dist_s1_pge.py
@@ -87,11 +87,9 @@ class DistS1PostProcessorMixin(PostProcessorMixin):
     _valid_layer_names = _main_output_layer_names + _confirmation_db_output_layer_names
 
     _product_id_pattern = (r'(?P<id>(?P<project>OPERA)_(?P<level>L3)_(?P<product_type>DIST(-ALERT)?)-(?P<source>S1)_'
-                           r'(?P<tile_id>T[^\W_]{5})_(?P<acquisition_ts>(?P<acq_year>\d{4})(?P<acq_month>\d{2})'
-                           r'(?P<acq_day>\d{2})T(?P<acq_hour>\d{2})(?P<acq_minute>\d{2})(?P<acq_second>\d{2})Z)_'
-                           r'(?P<creation_ts>(?P<cre_year>\d{4})(?P<cre_month>\d{2})(?P<cre_day>\d{2})T'
-                           r'(?P<cre_hour>\d{2})(?P<cre_minute>\d{2})(?P<cre_second>\d{2})Z)_(?P<sensor>S1[AC]?)_'
-                           r'(?P<spacing>30)_(?P<product_version>v\d+[.]\d+))')
+                           r'(?P<tile_id>T[^\W_]{5})_(?P<acquisition_ts>\d{4}\d{2}\d{2}T\d{2}\d{2}\d{2}Z)_'
+                           r'(?P<creation_ts>\d{4}\d{2}\d{2}T\d{2}\d{2}\d{2}Z)_(?P<sensor>S1[AC]?)_(?P<spacing>30)_'
+                           r'(?P<product_version>v\d+[.]\d+))')
 
     _granule_filename_pattern = (_product_id_pattern + rf'((_(?P<layer_name>{"|".join(_valid_layer_names)}))|'
                                                        r'_BROWSE)?[.](?P<ext>tif|tiff|png)$')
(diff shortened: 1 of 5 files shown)


src/opera/pge/dist_s1/dist_s1_pge.py
```diff
@@ -345,6 +346,23 @@ def _collect_dist_s1_product_metadata(self, geotiff_product):
             'spacing': 30  # meters/pixel
         }
 
+        # TODO: Replace these with metadata values sourced from the granule when available
+        #  (Or remove entirely if possible to refer to them straight through the MP dict)
+
+        match_result = self._granule_filename_re.match(basename(geotiff_product))
+
+        if match_result is None:
+            # This really should not happen due to bassing the _validate_outputs function but check anyway
```
collinss-jpl, line 355: bassing -> passing

src/opera/pge/dist_s1/dist_s1_pge.py
```diff
@@ -85,33 +86,33 @@ class DistS1PostProcessorMixin(PostProcessorMixin):
...
     _valid_layer_names = _main_output_layer_names + _confirmation_db_output_layer_names
 
-    def _validate_outputs(self):
-        product_id_pattern = (r'(?P<id>(?P<project>OPERA)_(?P<level>L3)_(?P<product_type>DIST(-ALERT)?)-(?P<source>S1)_'
-                              r'(?P<tile_id>T[^\W_]{5})_(?P<acquisition_ts>(?P<acq_year>\d{4})(?P<acq_month>\d{2})'
-                              r'(?P<acq_day>\d{2})T(?P<acq_hour>\d{2})(?P<acq_minute>\d{2})(?P<acq_second>\d{2})Z)_'
-                              r'(?P<creation_ts>(?P<cre_year>\d{4})(?P<cre_month>\d{2})(?P<cre_day>\d{2})T'
-                              r'(?P<cre_hour>\d{2})(?P<cre_minute>\d{2})(?P<cre_second>\d{2})Z)_(?P<sensor>S1[AC]?)_'
-                              r'(?P<spacing>30)_(?P<product_version>v\d+[.]\d+))')
+    _product_id_pattern = (r'(?P<id>(?P<project>OPERA)_(?P<level>L3)_(?P<product_type>DIST(-ALERT)?)-(?P<source>S1)_'
```
collinss-jpl, line 89: I just realized that we can simplify this regex a bit, since I don't think we'll have a need for granular access to the parsed timestamps, meaning we could remove the `?<acq_hour/minute/second>` and `?P<cre_hour/minute/second>` and just keep the full capture for `?p<acquisition_ts>` and `?p<creation_ts>`

collinss-jpl (approved):
Verified fix in the sandbox pipeline, so I think this is good to go. Just found one typo and a suggestion to simplify the regex we're using.

Also just FYI, you don't have to utilize the sandbox pipeline to run the integration test for a branch. The actual Integration Test pipeline can (and should) be used, since it will make it easier to find results from any given run (rather than having to remember and specify a specific build number)

RKuttruff:
> Also just FYI, you don't have to utilize the sandbox pipeline to run the integration test for a branch. The actual Integration Test pipeline can (and should) be used, since it will make it easier to find results from any given run (rather than having to remember and specify a specific build number)

Thanks, I mistakenly thought it was filtered to release/* branches like the release pipeline

src/opera/pge/dist_s1/dist_s1_pge.py
RKuttruff: Trimmed it down quite nicely, thanks

PR #646, opened 2025-04-21
SAS Log Traceback Parsing Fix

### LINKED ISSUE #645: [Bug]: Traceback capture reports entire input string
### Checked for duplicates

Yes - I've already checked

### Describe the bug

When a failure occurs during SAS execution, the PGE attempts to isolate the Traceback stack from the captured SAS log output via regular expression. However, when a Traceback is captured, the entire input string (the SAS log contents) are reported to the PGE log, rather than just the isolated Traceback.

### What did you expect?

When utilizing the `re.Match` object returned from the regular expression search, the `Match.start()` and `Match.end()` methods should be utilized to only report the matching substring (the captured Traceback) to the PGE log.

### Reproducible steps

```bash
1.
2.
3.
...
```

### Environment

```bash
...
```

collinss-jpl:
## Description
- This branch adds a fix to ensure that when a SAS failure occurs, only the Traceback parsed from the SAS log is included in the PGE log and what is reported to an SDS operator. Previously, the entire SAS log was mistakenly included instead of only the region matched by regular expression.

## Affected Issues
- Fixes #645 

## Testing
- Added a unit test for `run_utils.get_traceback_from_log()` to confirm fix.

### DIFF
--- a/src/opera/test/data/test_sas_log_with_traceback.txt
+++ b/src/opera/test/data/test_sas_log_with_traceback.txt
@@ -1,0 +1,128 @@
+<Example SAS output follows>
+Despeckling batch:   0%|          | 0/1 [00:00<?, ?it/s]
+Despeckling:   0%|          | 0/74 [00:00<?, ?it/s][A
+Despeckling:   1%|▏         | 1/74 [00:06<07:30,  6.18s/it][A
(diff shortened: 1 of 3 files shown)


RKuttruff (approved):
Looks good!

PR #649, opened 2025-04-24
Rework RTC homogeneity check to only check current burst set

### LINKED ISSUE #648: Update DIST-S1 PGE Input RTC Validation
Per a conversation with ADT: the baseline RTC set (post co-/cross-pol) can be heterogenous (ie, from different S1 satellites) while the current set (pre co-/cross-pol) must be homogenous (all from the same Sentinel sensor). The current set of PGE input validations assume all input RTCs needed to be homogenous, so they will need to be reworked to only check the current set.

RKuttruff:
## Description
- Reworks the DIST-S1 input validations to permit baseline bursts from different S1 sensors per discussion with ADT.

## Affected Issues
- Closes #648 

## Testing
- Reworked test case to inject S1C granule into current set instead of baseline set.
- Ensured tests are still passing on local & dev

### DIFF
--- a/src/opera/pge/dist_s1/dist_s1_pge.py
+++ b/src/opera/pge/dist_s1/dist_s1_pge.py
@@ -87,7 +87,7 @@ class DistS1PreProcessorMixin(PreProcessorMixin):
                 msg
             )
 
-    def __validate_rtc_filenames(self, all_rtcs):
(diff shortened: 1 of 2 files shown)


collinss-jpl (approved):
Very nice, way to stay out ahead of this @RKuttruff!

PR #650, opened 2025-04-24
TROPO Input/Output Product Validation

### LINKED ISSUE #641: Input/Output Product Validation for TROPO
The goal of this ticket is to add implementations to the TROPO pre- and post-processor mixins that validate the existence and validity of the input and output products to the TROPO SAS.

In the pre-processor mixin class, we need to implement an input validation method that examines the following portion(s) of the TROPO RunConfig:

- `input_file.input_file_path`

For the input file path listed in the above, we want to validate the following:
 
- File location exists
- File is non-zero in size
- File has an expected extension (.nc or .h5)

The util.input_validation.check_input function should be reused to accomplish these checks.

In the post-processor mixin class, we need a similar validation method, but for the output products. Based on the current delivery, we can expect the SAS to create the following output products for a single granule:

- `OPERA_L4_TROPO-ZENITH_<weather model time>_<production time>_HRES_v0.2.nc`
- `OPERA_L4_TROPO-ZENITH_<weather model time>_<production time>_HRES_v0.2.png`

The output validation method needs to confirm the following for each expected output file:

- File is non-zero in size
- File has expected extension (.nc or .png)
- File name conforms to the expected file name convention

Consult the CalVal product spec (available in Artifactory) for the latest file name convention.

Lastly, add two unit tests to the suite for TROPO that exercise the input/output validation functions in such a way that triggers each of the validation conditions listed.

kevinmarlis:
## Description
- Adds input/output product validation logic to `tropo_pge.py`
- Updates testing suite in `test_tropo_pge.py` to test against invalid input and output products along with the required updates to `test_tropo_config.yaml`

## Affected Issues
- Closes #641 

## Testing
- Tests ran successfully on opera-dev-pge

### DIFF
--- a/src/opera/pge/tropo/tropo_pge.py
+++ b/src/opera/pge/tropo/tropo_pge.py
@@ -165,9 +165,11 @@ class TROPOPostProcessorMixin(PostProcessorMixin):
             Any keyword arguments needed by the post-processor
         """
         
+        print(f'Running postprocessor for {self._post_mixin_name}')
+
+        self._run_sas_qa_executable()
         self._validate_outputs()
-        
-        super().run_postprocessor(**kwargs)
+        self._stage_output_files()
         
 
 
--- a/src/opera/pge/tropo/tropo_pge.py
+++ b/src/opera/pge/tropo/tropo_pge.py
@@ -46,13 +46,14 @@ class TROPOPreProcessorMixin(PreProcessorMixin):
         """
         super().run_preprocessor(**kwargs)
         
-        check_input(
-            self.runconfig.sas_config["input_file"]["input_file_path"], 
-            self.logger, 
-            self.runconfig.pge_name, 
-            valid_extensions=(".nc",), 
-            check_zero_size=True
-        )
+        for input_file in self.runconfig.input_files:
+            check_input(
+                input_file,
+                self.logger, 
+                self.runconfig.pge_name, 
+                valid_extensions=(".nc",), 
+                check_zero_size=True
+            )
 
 
 class TROPOPostProcessorMixin(PostProcessorMixin):
@@ -72,12 +73,15 @@ class TROPOPostProcessorMixin(PostProcessorMixin):
         
     def _validate_outputs(self):
         """
+        Confirms that there exists one and only of each expected output type.
+        Also confirms that for each output:
+        - File has nonzero size
+        - Filename matches expected pattern
         """
-
         filename_pattern = re.compile(
             rf"^OPERA_L4_TROPO-ZENITH_\d{{8}}T\d{{6}}Z_\d{{8}}T\d{{6}}Z_HRES_{re.escape('v' + self.SAS_VERSION)}$"
         )
-        
+
         output_product_path = abspath(self.runconfig.output_product_path)
         output_product_files = listdir(output_product_path)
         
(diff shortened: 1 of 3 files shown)


src/opera/pge/tropo/tropo_pge.py
```diff
@@ -50,11 +63,60 @@ class TROPOPostProcessorMixin(PostProcessorMixin):
...
+                error_msg = f"Could not locate {filename_ext} file."
+                self.logger.critical(self.name, ErrorCode.OUTPUT_NOT_FOUND, error_msg)
+                
+            # Check for second file, if found raise error
+            try:
+                next(gen)
+                error_msg = f"Found incorrect number of {filename_ext} files."
+                self.logger.critical(self.name, ErrorCode.INVALID_OUTPUT, error_msg)
+            except StopIteration:
+                pass
```
collinss-jpl, line 103: Nice approach here, very Pythonic

src/opera/pge/tropo/tropo_pge.py
```diff
@@ -50,11 +63,60 @@ class TROPOPostProcessorMixin(PostProcessorMixin):
...
+            try:
+                output_filename = next(gen)
+            except StopIteration:
+                error_msg = f"Could not locate {filename_ext} file."
+                self.logger.critical(self.name, ErrorCode.OUTPUT_NOT_FOUND, error_msg)
+                
+            # Check for second file, if found raise error
+            try:
+                next(gen)
+                error_msg = f"Found incorrect number of {filename_ext} files."
```
collinss-jpl, line 100: Just curious, can we still use `len(gen)` here to include the number of actual files found in the error message, or has use of `next` already invalidated that somehow?

src/opera/pge/tropo/tropo_pge.py
```diff
@@ -96,14 +157,18 @@ def run_postprocessor(self, **kwargs):
...
         Parameters
         ----------
         **kwargs: dict
             Any keyword arguments needed by the post-processor
         """
+        
+        self._validate_outputs()
+        
         super().run_postprocessor(**kwargs)
+
```
collinss-jpl, line 171: This happens to work alright for TROPO, but for other PGEs we typically forgo the call to `super().run_postprocessor()` in favor of recomposing the steps here so we have full control over the ordering. So I would rather see this:

```
print(f'Running postprocessor for {self._post_mixin_name}')

self._run_sas_qa_executable()
self._validate_outputs()
self._stage_output_files()
```

This also ensure the print statement occurs where we expect it to (prior to any validation checks).

collinss-jpl (changes requested):
Looks great @kevinmarlis, just one little quibble that should be easy to address

src/opera/pge/tropo/tropo_pge.py
kevinmarlis: Because `gen` is of type `generator` you can't determine size without exhausting the generator. This generator approach is maybe too clever/convoluted for the need. I could switch to a list / size of list approach if you think that would be more direct.

src/opera/pge/tropo/tropo_pge.py
kevinmarlis: Got it, thanks! This is what @RKuttruff suggested on Slack - I figured the next round of additions would necessitate recomposing the steps so I didn't include it here, but I'll add it in now.

src/opera/pge/tropo/tropo_pge.py
collinss-jpl: I think what's here is fine for now

collinss-jpl (approved):

PR #670, opened 2025-05-15
DISP-NI Input/Output validations

### LINKED ISSUE #669: Input/Output Product Validation for DISP-NI
The goal of this ticket is to implement the input and output product validation steps for the DISP-NI PGE.

DISP-NI shares a similar interface with DISP-S1, so an additional goal of this ticket will be to leverage (or retrofit) the existing DISP-S1 validation code wherever possible. If feasible, the pre- and post-processor mixins for DISP-NI can inherit from the DISP-S1 counterparts.

For the input product validation step, we should be able to reuse `input_validation.validate_disp_inputs()` by making some minor tweaks to have it include checks on runconfig fields that specific to DISP-NI (such as `gunw_files`, or `frame_to_bounds_json`).

For validating the algorithm parameters config, we should be able to reuse `input_validation.validate_algorithm_parameters_config()` directly.

For output product validation, `DispS1PostProcessorMixin._validate_output()` seems to be just about where we need it, but the conditional on line 212 will need a slight adjustment to make it work for both DISP-S1 and NI.

As with the other validation tasks, there should also be unit tests for input validation, algorithm parameters validation and output validation.

RKuttruff:
## Description
- This PR reuses the DISP-S1 input and output validations for the DISP-NI product

## Affected Issues
- Closes #669 

## Testing
- Fixed existing unit tests to work with validations enabled
- Added new unit tests for input & output validations

### DIFF
--- a/src/opera/util/input_validation.py
+++ b/src/opera/util/input_validation.py
@@ -307,23 +307,30 @@ def validate_disp_inputs(runconfig, logger, name):
     dyn_anc_file_group = runconfig.sas_config['dynamic_ancillary_file_group']
     static_anc_file_group = runconfig.sas_config['static_ancillary_file_group']
 
-    check_input_list(input_file_group['cslc_file_list'], logger, name,
-                     valid_extensions=('.h5',), check_zero_size=True)
-
-    cslc_burst_id_set = get_cslc_input_burst_id_set(
-        runconfig.sas_config['input_file_group']['cslc_file_list'],
-        logger, name
-    )
-
-    if ('static_layers_files' in dyn_anc_file_group and
-            isinstance(dyn_anc_file_group['static_layers_files'], list) and
-            len(dyn_anc_file_group['static_layers_files']) > 0):
-        check_input_list(dyn_anc_file_group['static_layers_files'], logger, name,
+    if runconfig.pge_name == 'DISP_S1_PGE':
+        check_input_list(input_file_group['cslc_file_list'], logger, name,
                          valid_extensions=('.h5',), check_zero_size=True)
-        check_disp_s1_ancillary_burst_ids(cslc_burst_id_set,
-                                          dyn_anc_file_group['static_layers_files'],
-                                          logger,
-                                          name)
+
+        cslc_burst_id_set = get_cslc_input_burst_id_set(
+            runconfig.sas_config['input_file_group']['cslc_file_list'],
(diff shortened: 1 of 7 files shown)


src/opera/util/input_validation.py
```diff
@@ -307,23 +307,30 @@ def validate_disp_inputs(runconfig, logger, name):
     dyn_anc_file_group = runconfig.sas_config['dynamic_ancillary_file_group']
     static_anc_file_group = runconfig.sas_config['static_ancillary_file_group']
 
-    check_input_list(input_file_group['cslc_file_list'], logger, name,
-                     valid_extensions=('.h5',), check_zero_size=True)
+    if runconfig.pge_name == 'DISP_S1_PGE':
```
collinss-jpl, line 310: Approach looks good here, but the docstring should be updated to indicate that this function is now intended for use with two PGEs

src/opera/util/input_validation.py
```diff
@@ -307,23 +307,30 @@ def validate_disp_inputs(runconfig, logger, name):
...
+    else:
+        check_input_list(input_file_group['gslc_file_list'], logger, name,
                          valid_extensions=('.h5',), check_zero_size=True)
-        check_disp_s1_ancillary_burst_ids(cslc_burst_id_set,
-                                          dyn_anc_file_group['static_layers_files'],
-                                          logger,
-                                          name)
+
+        # TODO: Is there a NI equivalent to the burst set - ancillaries validation of DISP-S1?
+        #  I don't seem to see a burst ID in the input filenames
```
collinss-jpl, line 333: Had a look at the sample GSLC and GUNW products we got from the interface delivery, and while I did not see anything that could be correlated in the filenames, we could at the very least ensure the runconfig specifies the same number of each file. We'll need to double-check with ADT that this is a true invariant though.

RKuttruff: Should we do that in this PR or check with ADT first?

src/opera/util/input_validation.py
RKuttruff: Docstring updated

src/opera/util/input_validation.py
collinss-jpl: Lets go ahead and add it now

collinss-jpl (approved):

PR #671, opened 2025-05-15
TROPO ISO XML Generation

### LINKED ISSUE #654: ISO XML and Measured Parameters Config for TROPO
This ticket is to add support for generation of the ISO XML output product for the TROPO PGE.

At a high level, this will include:

- [x] Creation of the ISO XML Jinja2 template for use with TROPO
- [x] Derivation of the current metadata available within a TROPO CalVal product (Consult the [get_hdf5_group_as_dict](https://github.com/nasa/opera-sds-pge/blob/main/src/opera/util/h5_utils.py#L39) function)
- [x] Based on the available metadata and the documentation from the latest product spec, derive the Measured Parameters config file for use with the ISO XML template instantiation
- [x] Implement the `h5_utils.get_tropo_product_metadata()` function that uses `get_hdf5_group_as_dict` to extract the relevant metadata from a TROPO product (we want to avoid extracting any raster/image layers to save on memory)
- [x] Implement the `TROPOPostProcessorMixin._collect_tropo_metadata()` method that utilizes `h5_utils.get_tropo_product_metadata()` to extract metadata from a product, then creates the Measured Parameters dictionary for use with ISO XML template (see [utils.render_jinja2.augment_hdf5_measured_parameters](https://github.com/nasa/opera-sds-pge/blob/main/src/opera/util/render_jinja2.py#L289))
- [x] Implement `TROPOPostProcessorMixin._create_iso_metadata()` which is responsible for invoking `render_jinja2` using the template and the dictionary of collected metadata.
- [x] Implement `TROPOPostProcessorMixin._stage_output_files()`, which performs the same tasks as the base PGE implementation, but uses `TROPOPostProcessorMixin._collect_tropo_metadata()` to gather the input metadata to feed to `_create_iso_metadata()`
- [x] Add a unit test to test_tropo.py that exercises ISO XML creation, and ensures that all placeholders have been filled as expected

kevinmarlis:
## Description
- Adds support for ISO XML and Measured Parameters Config for TROPO PGE
- Ensures that framework for populating ISO XML with TROPO output product metadata works as intended.
- There are a few outstanding values that are using placeholders until we get confirmation otherwise:
    - latitude / longitude spacing: Product spec says ~0.07 degree resolution which is equivalent to roughly 8000 meters
    - spatial extent: Product is global on a regular 2d grid, so we're using the min/max lat/lon values to define the spatial extent
    - zone identifier: No reference in product spec, but the product is global
    - start_time / end_time: Single time value per product along with a temporal_resolution metadata field of "6h". Currently using time value from product for start_time and start_time + timedelta(hours=6) for end_time

## Affected Issues
- Closes #654 

## Testing
- Added ISO XML test which ensures all placeholders have been filled and that the appropriate error is raised when unable to generate ISO XML
- Will update after running Build/Test Jenkins pipeline

### DIFF
--- a/src/opera/pge/tropo/templates/OPERA_ISO_metadata_L4_TROPO_template.xml.jinja2
+++ b/src/opera/pge/tropo/templates/OPERA_ISO_metadata_L4_TROPO_template.xml.jinja2
@@ -493,21 +493,21 @@
                     <gmd:geographicElement>
                         <gmd:EX_GeographicBoundingBox>
                             <gmd:westBoundLongitude>
-                                <gco:Decimal>{{ product_output.geospatial_lon_min }}{# doc ID ISO_OPERA_westBoundLongitude #}</gco:Decimal>
+                                <gco:Decimal>{{ product_output.spatial_extent.lon_min }}{# doc ID ISO_OPERA_westBoundLongitude #}</gco:Decimal>
                             </gmd:westBoundLongitude>
                             <gmd:eastBoundLongitude>
-                                <gco:Decimal>{{ product_output.geospatial_lon_max }}{# doc ID ISO_OPERA_eastBoundLongitude #}</gco:Decimal>
+                                <gco:Decimal>{{ product_output.spatial_extent.lon_max }}{# doc ID ISO_OPERA_eastBoundLongitude #}</gco:Decimal>
                             </gmd:eastBoundLongitude>
                             <gmd:southBoundLatitude>
-                                <gco:Decimal>{{ product_output.geospatial_lat_min }}{# doc ID ISO_OPERA_southBoundLatitude #}</gco:Decimal>
+                                <gco:Decimal>{{ product_output.spatial_extent.lat_min }}{# doc ID ISO_OPERA_southBoundLatitude #}</gco:Decimal>
                             </gmd:southBoundLatitude>
                             <gmd:northBoundLatitude>
-                                <gco:Decimal>{{ product_output.geospatial_lat_max }}{# doc ID ISO_OPERA_northBoundLatitude #}</gco:Decimal>
+                                <gco:Decimal>{{ product_output.spatial_extent.lat_max }}{# doc ID ISO_OPERA_northBoundLatitude #}</gco:Decimal>
                             </gmd:northBoundLatitude>
                         </gmd:EX_GeographicBoundingBox>
                     </gmd:geographicElement>
                     <!-- This section documents the ZoneIdentifier -->
-                    <gmd:geographicElement>
+                    {# <gmd:geographicElement>
                         <gmd:EX_GeographicDescription id="ZoneIdentifier">
                             <gmd:geographicIdentifier>
                                 <gmd:MD_Identifier>
@@ -523,15 +523,15 @@
                                 </gmd:MD_Identifier>
                             </gmd:geographicIdentifier>
                         </gmd:EX_GeographicDescription>
-                    </gmd:geographicElement>
+                    </gmd:geographicElement> #}
                     <!-- This is the granules temporal extent -->
                     <gmd:temporalElement>
                         <!--RangeDateTime-->
                         <gmd:EX_TemporalExtent id="boundingTemporalExtent">
                             <gmd:extent>
-                                <gml:TimePeriod gml:id="DIST_Time_Range">
-                                    <gml:beginPosition>{{ product_output.acquisition_start_time }}{# ISO_OPERA_refDateTime #}</gml:beginPosition>
-                                    <gml:endPosition>{{ product_output.acquisition_end_time }}{# ISO_OPERA_secDateTime #}</gml:endPosition>
+                                <gml:TimePeriod gml:id="TROPO_Time_Range">
+                                    <gml:beginPosition>{{ product_output.temporal_extent.start_time }}{# ISO_OPERA_refDateTime #}</gml:beginPosition>
+                                    <gml:endPosition>{{ product_output.temporal_extent.end_time }}{# ISO_OPERA_secDateTime #}</gml:endPosition>
                                 </gml:TimePeriod>
                             </gmd:extent>
                         </gmd:EX_TemporalExtent>
--- a/src/opera/pge/tropo/templates/OPERA_ISO_metadata_L4_TROPO_template.xml.jinja2
+++ b/src/opera/pge/tropo/templates/OPERA_ISO_metadata_L4_TROPO_template.xml.jinja2
@@ -507,12 +507,12 @@
                         </gmd:EX_GeographicBoundingBox>
                     </gmd:geographicElement>
                     <!-- This section documents the ZoneIdentifier -->
-                    {# <gmd:geographicElement>
+                    <gmd:geographicElement>
                         <gmd:EX_GeographicDescription id="ZoneIdentifier">
                             <gmd:geographicIdentifier>
                                 <gmd:MD_Identifier>
                                     <gmd:code>
-                                        <gco:CharacterString>{{ product_output.zoneIdentifier }}{# ISO_OPERA_zoneIdentifier #}</gco:CharacterString>
+                                        <gco:CharacterString>{{ product_output.zone_identifier }}{# ISO_OPERA_zoneIdentifier #}</gco:CharacterString>
                                     </gmd:code>
                                     <gmd:codeSpace>
                                         <gco:CharacterString>gov.nasa.esdis.umm.zoneidentifier</gco:CharacterString>
@@ -523,7 +523,7 @@
                                 </gmd:MD_Identifier>
                             </gmd:geographicIdentifier>
                         </gmd:EX_GeographicDescription>
-                    </gmd:geographicElement> #}
+                    </gmd:geographicElement>
                     <!-- This is the granules temporal extent -->
                     <gmd:temporalElement>
                         <!--RangeDateTime-->
--- a/src/opera/pge/tropo/templates/OPERA_ISO_metadata_L4_TROPO_template.xml.jinja2
+++ b/src/opera/pge/tropo/templates/OPERA_ISO_metadata_L4_TROPO_template.xml.jinja2
@@ -117,7 +117,7 @@
                     <gmd:authority>
                         <gmd:CI_Citation>
                             <gmd:title>
-                                <gco:CharacterString>Universal Transverse Mercator (UTM)</gco:CharacterString>
+                                <gco:CharacterString>World Geodetic System 1984 (WGS84)</gco:CharacterString>
                             </gmd:title>
                         </gmd:CI_Citation>
                     </gmd:authority>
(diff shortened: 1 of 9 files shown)


RKuttruff:
@kevinmarlis Secrets detection is flagging this PR, which is normal for ISO XML additions because they contain contact info. I'll message you with the details

src/opera/pge/tropo/templates/OPERA_ISO_metadata_L4_TROPO_template.xml.jinja2
```diff
@@ -0,0 +1,720 @@
...
+        </gmd:MD_GridSpatialRepresentation>
+    </gmd:spatialRepresentationInfo>
+    <gmd:referenceSystemInfo>
+        <gmd:MD_ReferenceSystem>
+            <gmd:referenceSystemIdentifier>
+                <gmd:MD_Identifier>
+                    <gmd:authority>
+                        <gmd:CI_Citation>
+                            <gmd:title>
+                                <gco:CharacterString>Universal Transverse Mercator (UTM)</gco:CharacterString>
```
RKuttruff, line 120: The product spec says this is in WGS84 reference system. I don't know if that requires changing this or how to properly cite for WGS84

src/opera/util/h5_utils.py
```diff
@@ -141,6 +141,92 @@ def convert_h5py_dataset(dataset_object):
     return result
 
 
+def get_hdf5_attrs_as_dict(file_name, group_path, ignore_keys=[]):
```
collinss-jpl, line 144: Usually want to avoid assigning a mutable object as a default argument value. `ignore_keys` should default to `None` and this function should initialize it to an empty list if `None` is provided.

src/opera/pge/tropo/templates/OPERA_ISO_metadata_L4_TROPO_template.xml.jinja2
```diff
@@ -0,0 +1,720 @@
...
+            <gmd:axisDimensionProperties>
+                <gmd:MD_Dimension>
+                    <gmd:dimensionName>
+                        <gmd:MD_DimensionNameTypeCode codeList="https://cdn.earthdata.nasa.gov/iso/resources/Codelist/gmxCodelists.xml#MD_DimensionNameTypeCode" codeListValue="row">row</gmd:MD_DimensionNameTypeCode>
+                    </gmd:dimensionName>
+                    <gmd:dimensionSize>
+                        <gco:Integer>{{ product_output.xCoordinates.size }}{# ISO_OPERA_rangeCount #}</gco:Integer>
+                    </gmd:dimensionSize>
+                    <gmd:resolution>
+                        <gco:Length uom="meter">{{ product_output.xCoordinates.spacing }}{# ISO_OPERA_rangePixelSize #}</gco:Length>
```
RKuttruff, line 88: I couldn't find it in the schema on a quick search, but I wonder if this can be given in degrees instead of meters

collinss-jpl (changes requested):
Branch is looking good, just a few minor things to address

src/opera/pge/tropo/templates/OPERA_ISO_metadata_L4_TROPO_template.xml.jinja2
```diff
@@ -0,0 +1,720 @@
...
+                <gmd:MD_CharacterSetCode codeList="http://www.ngdc.noaa.gov/metadata/published/xsd/schema/resources/Codelist/gmxCodelists.xml#MD_CharacterSetCode" codeListValue="utf8">utf8</gmd:MD_CharacterSetCode>
+            </gmd:characterSet>
+            <gmd:topicCategory>
+                <gmd:MD_TopicCategoryCode codeList="https://cdn.earthdata.nasa.gov/iso/resources/Codelist/gmxCodelists.xml#MD_TopicCategoryCode" codeListValue="geoscientificInformation">geoscientificInformation</gmd:MD_TopicCategoryCode>
+            </gmd:topicCategory>
+            <gmd:environmentDescription>
+                <gco:CharacterString>Data product generated in Network Common Data From version 4 format with ISO 19115 conformant metadata.</gco:CharacterString>
+            </gmd:environmentDescription>
+            <!-- This section documents the granules spatial and temporal extent as well as the grid mapping names and the projection names - NOT the TilingIdentificationSystem. -->
+            <gmd:extent>
```
RKuttruff, line 489: Should we include vertical extent here since this product exists at multiple elevation layers?

src/opera/pge/tropo/templates/OPERA_ISO_metadata_L4_TROPO_template.xml.jinja2
```diff
@@ -0,0 +1,720 @@
...
+    </gmd:contentInfo>
+    <!-- This section holds the Related URLs that pertain to distributions - where UMM-G RelatedUrl/Type = GET SERVICE, GET DATA, OPENDAP DATA ACCESS -->
+    <gmd:distributionInfo>
+        <gmd:MD_Distribution>
+            <gmd:distributor>
+                <gmd:MD_Distributor>
+                    <gmd:distributionFormat>
+                        <gmd:MD_Format>
+                            <gmd:name>
+                                <gco:CharacterString>GeoTIFF</gco:CharacterString>
```
RKuttruff, line 596: This should be NetCDF

src/opera/pge/tropo/tropo_pge.py
```diff
@@ -150,7 +152,254 @@ def _checksum_output_products(self):
...
+            Dictionary containing the custom metadata as expected by the ISO
+            metadata Jinja2 template.
+
+        """
+        custom_metadata = {
+            'ISO_OPERA_FilePackageName': granule_filename,
+            'ISO_OPERA_ProducerGranuleId': granule_filename,
+            'MetadataProviderAction': "creation",
+            'GranuleFilename': granule_filename,
+            'ISO_OPERA_ProjectKeywords': ['OPERA', 'JPL', 'DSWx', 'Dynamic', 'Surface', 'Water', 'Extent'],
```
RKuttruff, line 278: Should update these to apply to TROPO

src/opera/test/pge/tropo/test_tropo_pge.py
```diff
@@ -121,6 +126,16 @@ def test_tropo_pge_execution(self):
...
+        
+        # Check that the catalog metadata file was created in the output directory
+        expected_catalog_metadata_file = join(
+            pge.runconfig.output_product_path, pge._catalog_metadata_filename())
+        self.assertTrue(os.path.exists(expected_catalog_metadata_file))
+
+        # # Check that the ISO metadata file was created and filled in as expected
+        expected_iso_metadata_file = join(
+            pge.runconfig.output_product_path, pge._iso_metadata_filename())
+        self.assertTrue(os.path.exists(expected_iso_metadata_file))
```
RKuttruff, line 138: You should add a check here that the ISO file is fully populated and does not contain the template undefined error string
Reference: https://github.com/nasa/opera-sds-pge/blob/f13869f9c67a1b3451c3ca75fe10bcfcb09d96dc/src/opera/test/pge/dist_s1/test_dist_s1_pge.py#L179-L182

RKuttruff (changes requested):
I went over the ISO template and raised some questions, but also found some things to change/add

src/opera/pge/tropo/tropo_pge.py
```diff
@@ -150,7 +152,254 @@ def _checksum_output_products(self):
...
+    def _stage_output_files(self):
+        # Gather the list of output files produced by the SAS
+        output_products = self.runconfig.get_output_product_filenames()
+
+        # For each output file name, assign the final file name matching the
+        # expected conventions. Also extracts netCDF output product.
+        nc_product = None
+        for output_product in output_products:
+            if output_product.endswith('nc'):
+                nc_product = output_product
```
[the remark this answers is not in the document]
kevinmarlis: Even though we're not currently doing any file renaming for TROPO, this for loop does include that step. I could break out the '.nc' search into it's own for loop and add the break.

src/opera/util/h5_utils.py
```diff
@@ -141,6 +141,92 @@ def convert_h5py_dataset(dataset_object):
...
+    -------
+    extent : (float, float, float, float)
+        Tuple containing (min_lon, max_lon, min_lat, max_lat).
+
+    """
+    with h5py.File(file_name, 'r') as h5file:
+        group_object = h5file.get(group_path)
+
+        if group_object is None:
+            raise RuntimeError(f"An error occurred retrieving object '{group_path}' "
```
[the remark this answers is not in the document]
kevinmarlis: I'll go ahead and add it to `get_hdf5_group_as_dict` as well since I'm in here already

src/opera/pge/tropo/templates/OPERA_ISO_metadata_L4_TROPO_template.xml.jinja2
kevinmarlis: We can, but the extent is variable as the vertical layers produced are configurable via the runconfig.

src/opera/pge/tropo/templates/OPERA_ISO_metadata_L4_TROPO_template.xml.jinja2
kevinmarlis: I think we can use degree but would need to change Length to Angle as well. I can try and track down some more resources on this.

src/opera/pge/tropo/templates/OPERA_ISO_metadata_L4_TROPO_template.xml.jinja2
kevinmarlis: Ah, good catch! I believe it should be 
```<gco:CharacterString>World Geodetic System 1984 (WGS84)</gco:CharacterString>```

src/opera/pge/tropo/templates/OPERA_ISO_metadata_L4_TROPO_template.xml.jinja2
RKuttruff: FWIW I tried pulling the ISO files for global WGS84 collections like MUR, and they didn't even have some of these sections like spacing and zone identifier

src/opera/util/h5_utils.py
kevinmarlis: Great catch, thank you!

kevinmarlis:
Addressed concerns and closed conversations where changes were implemented. Test is passing on opera-dev-pge. Outstanding items are:

- ISO XML WGS84 usage of spacing and zone identifier
- ISO XML addition of vertical extent

RKuttruff (commented):
All my main requested changes have been resolved. So I think this is good to go pending discussion on the outstanding points/questions Kevin mentioned.

collinss-jpl (approved):

PR #692, opened 2025-06-17
TROPO Ancillary Output Filename Support

### LINKED ISSUE #687: [Bug]: TROPO PGE does not apply correct file name coventions to ISO XML product
### Checked for duplicates

Yes - I've already checked

### Describe the bug

In the TROPO PGE workflow, the SAS is responsible for assigning the correct file names to the .nc and .png products based on the official conventions. However, we have neglected to apply the same conventions to the ISO.xml product that the PGE produces.

The simplest solution might be to just cache the output filename of the .nc (or .png) product from the SAS, then just replace the existing file extension with `.iso.xml`

### What did you expect?

TROPO PGE produces an .iso.xml product with the correct naming conventions (same as output products).

### Reproducible steps

```bash
1.
2.
3.
...
```

### Environment

```bash
...
```

sdcolli1:
## Description
- This branch adds support for application of the TROPO output file name conventions to the ancillary outputs created by the PGE (.iso.xml, .log, etc).
- There are also some small tweaks to the unit test suite so the dummy output products are created during PGE execution, rather than during setup of the test suite. 

## Affected Issues
- Fixes #687

## Testing
- All unit tests are passing
- Integration test has been run for this branch in Jenkins, and all product comparisons are passing

### DIFF
--- a/src/opera/test/data/test_tropo_config.yaml
+++ b/src/opera/test/data/test_tropo_config.yaml
@@ -15,12 +15,15 @@ RunConfig:
       PrimaryExecutable:
         ProductIdentifier: TROPO
         ProductVersion: "0.2"
-        ProgramPath: /bin/echo
+        ProgramPath: mkdir
         ProgramOptions:
-        - '/bin/echo TROPO invoked with RunConfig'
+          - '-p tropo_pge_test/output_dir;'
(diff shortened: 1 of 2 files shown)


RKuttruff (approved):
Looks good!

PR #706, opened 2025-07-03
Add more DIST-S1 input validations

### LINKED ISSUE #695: [Feature]: Add new validations for DIST-S1 PGE input
The following validations should be added to the DIST-S1 PGE:
- [x] Input RTC lists should contain no duplicate products for each burst/sensing time (ie, no files with the same burst & sensing time but different production time)
- [x] Previous product validations:
  - [x] Files exist
  - [x] Files are not empty
  - [x] Files are named correctly
  - [x] Exactly a single, full set of layers or or nothing at all
- [x] Add/update unit tests

RKuttruff:
## Description
- This PR adds new input validations to the DIST-S1 PGE
- Validations:
  - No RTCs with duplicate burst IDs/acquisition times/polarities
  - Previous DIST product (if provided) is a list of extant, nonempty, validly named GeoTIFFs with all layers defined in the DIST-S1 spec

## Affected Issues
- Closes #695 

## Testing
- Added and updated unit tests
  - Passing on local & dev

### DIFF
--- a/src/opera/test/pge/dist_s1/test_dist_s1_pge.py
+++ b/src/opera/test/pge/dist_s1/test_dist_s1_pge.py
@@ -102,8 +102,8 @@ class DistS1PgeTestCase(unittest.TestCase):
             Product ID to create the output files for.
         band_data: tuple of str
             Files to add to the output directory.
-        directory: str
-            Directory to create the test files in.
+        directory: str (optional)
+            Directory to create the test files in. If not provided, the test case output directory is used.
         empty_file: bool
             if 'True' do not add text to the file (leave empty)
             if 'False' (default) add 'Test data string' to the file


src/opera/pge/dist_s1/dist_s1_pge.py
```diff
@@ -271,6 +331,61 @@ def _validate_rtcs(self):
...
+        Checks:
+        1. All files exist
+        2. All files are nonempty
+        3. All files are .tif files
+        4. Correct number of files (one for each TIFF layer) provided
+        5. All files are named conformant to DIST-S1 product specification
+        6. Exact set of layers provided (ie, no duplicates)
+        """
+        sas_config = self.runconfig.sas_config
+        previous_product = sas_config["run_config"].get("pre_dist_s1_product", [])
```
collinss-jpl, line 349: Its a bit confusing that `pre_dist_s1_product` is actually a list of products plural as defined by the schema. So I would rather see this attriubte named `previous_products` to make the following code clearer.

src/opera/pge/dist_s1/dist_s1_pge.py
```diff
@@ -271,6 +331,61 @@ def _validate_rtcs(self):
...
+
+        if len(previous_product) != len(geotiff_layer_names):
+            msg = f'Unexpected number of files in previous product: {len(previous_product)}'
+            self.logger.critical(
+                self.name,
+                ErrorCode.INVALID_INPUT,
+                msg
+            )
+
+        product_band_matches = [_granule_filename_re.match(basename(f)) for f in previous_product]
```
collinss-jpl, line 369: Another nitpick, but since we'll be changing `previous_product` to `previous_products`, I would also rather see `previous_product` used here instead of `f`. Best to avoid one-letter variables names as a general rule.

src/opera/pge/dist_s1/dist_s1_pge.py
```diff
@@ -295,6 +410,12 @@ def run_preprocessor(self, **kwargs):
         )
         self._validate_rtcs()
 
+        sas_config = self.runconfig.sas_config
+        previous_product = sas_config["run_config"].get("pre_dist_s1_product", [])
```
collinss-jpl, line 414: Same note here on `previous_product` as above

src/opera/test/pge/dist_s1/test_dist_s1_pge.py
```diff
@@ -100,6 +102,8 @@ def generate_band_data_output(self, product_id, band_data, empty_file=False, cle
             Product ID to create the output files for.
         band_data: tuple of str
             Files to add to the output directory.
+        directory: str
+            Directory to create the test files in.
```
collinss-jpl, line 106: This should note that this is an optional argument, and what the default is if not provided.

collinss-jpl (commented):
Excellent work on the additional test coverage. Only had some minor findings and suggestions.

src/opera/pge/dist_s1/dist_s1_pge.py
RKuttruff: It's a single product in the form of multiple .tif files, so maybe `previous_product_files`?

RKuttruff:
Thanks for the suggestions! Updates have been pushed.

PR #728, opened 2025-09-04
DISP-S1-STATIC Bounding Box Derived From opera_utils

### LINKED ISSUE #724: [New Feature]: Derive DISP-S1-STATIC Bounding Box using opera_utils
### Checked for duplicates

Yes - I've already checked

### Alternatives considered

Yes - and alternatives don't suffice

### Related problems

We currently have a [placeholder](https://github.com/nasa/opera-sds-pge/blob/main/src/opera/pge/disp_s1/disp_s1_pge.py#L1281-L1285) within the Post-processor of the DISP-S1-STATIC PGE that assigns zeros for the Geospatial bounding box we include in the ISO XML. This is because we've been expecting those values to come from the product metadata, but after talking with Scott S. he suggested the following workaround solution:

```python
epsg, bounds = opera_utils.get_frame_bbox(
  frame_id=self.runconfig.input_file_group.frame_id,
  json_file=self.runconfig.static_ancillary_file_group.frame_to_burst_json,
)
```

Documentation on opera_utils.get_frame_bbox can be found [here](https://github.com/opera-adt/opera-utils/blob/8ced3978cc6e00d3b8596fead54b1e4826ff4563/src/opera_utils/burst_frame_db.py#L277)

Note that opera_utils should already be installed within the environment of the SAS container. We should also be able to reference it within our requirements.txt to ensure its available in our dev environments.

### Describe the feature request

I would like to investigate implementing this approach so we can address the missing/zero bbox issue (which ASF.DAAC has noticed and asked about) without requiring a new SAS delivery. If this ends up introducing container build issues from the introduction of the opera_utils dependency on our code, then I want to know that sooner than later.

kevinmarlis:
## Description
- Makes use of `opera-utils` library for dynamically obtaining bounding box values used in iso metadata
- Adds `operautils` to `requirements.txt` for dev environment purposes
- Includes new `opera-s1-disp-frame-to-burst.json` file used in testing

## Affected Issues
- Closes #724 

## Testing
- Unit and int tests passing on opera-dev-pge
- iso.xml contains non-zero values for bounding rectangle:
```
                    <gmd:geographicElement>
                        <gmd:EX_GeographicBoundingBox>
                            <gmd:westBoundLongitude>
                                <gco:Decimal>523440.0</gco:Decimal>
                            </gmd:westBoundLongitude>
                            <gmd:eastBoundLongitude>
                                <gco:Decimal>811380.0</gco:Decimal>
                            </gmd:eastBoundLongitude>
                            <gmd:southBoundLatitude>
                                <gco:Decimal>4055790.0</gco:Decimal>
                            </gmd:southBoundLatitude>
                            <gmd:northBoundLatitude>
                                <gco:Decimal>4261770.0</gco:Decimal>
                            </gmd:northBoundLatitude>
                        </gmd:EX_GeographicBoundingBox>
                    </gmd:geographicElement>
```


src/opera/util/geo_utils.py
```diff
@@ -28,6 +30,22 @@
...
+# pylint: disable=import-error,invalid-name 
+try:
+    from opera_utils import get_frame_geodataframe
+    from shapely import union_all, affinity
+    from shapely.geometry import MultiPolygon
+except (ImportError, ModuleNotFoundError): # pragma: no cover
+    get_frame_geodataframe = MagicMock()           # pragma: no cover
+    union_all = MagicMock()             # pragma: no cover
+    affinity = MagicMock()              # pragma: no cover
+    MultiPolygon = MagicMock()          # pragma: no cover
```
RKuttruff, line 46: I don't think this is super critical, but `MagicMock()` can raise an error in `isinstance` calls

RKuttruff (commented):
Looks good except for one potential bug.

Unit tests passed on dev.

Tested and verified ISO extent set properly
```xml
<gmd:geographicElement>
    <gmd:EX_BoundingPolygon>
        <gmd:polygon>
            <gml:Polygon gml:id="DISP_Frame_Polygon">
                <gml:exterior>
                    <gml:LinearRing>
                        <gml:description>Bounding polygon of the displacement image</gml:description>
                        <gml:posList>(-122.68045 36.926455 -122.369757 38.457397 -119.502762 38.173828 -119.86972 36.651668 -122.68045 36.926455)</gml:posList>
                    </gml:LinearRing>
                </gml:exterior>
            </gml:Polygon>
        </gmd:polygon>
    </gmd:EX_BoundingPolygon>
</gmd:geographicElement>
```

src/opera/util/geo_utils.py
RKuttruff:
```
>       if isinstance(polygon, MultiPolygon):
E       TypeError: isinstance() arg 2 must be a type, a tuple of types, or a union
```

src/opera/util/geo_utils.py
```diff
@@ -196,3 +214,37 @@ def get_geographic_boundaries_from_mgrs_tile(mgrs_tile_name):
...
+        json_file=frame_geometries
+    )
+    
+    polygon = gdf_frames.loc[frame_id].geometry
+    
+    # Handles the case where polygon crosses the anti-meridian
+    # Translate the polygon with negative longitudes to be right of +180
+    if isinstance(polygon, MultiPolygon):
+        gg_neg, gg_pos = polygon.geoms
+        polygon = union_all([gg_pos, affinity.translate(gg_neg, 360)])
```
collinss-jpl, line 247: This might be overkill for what I was looking for. Are we certain we need it? 

Without it it seem like we wouldn't need the following imports
```
from shapely import union_all, affinity
from shapely.geometry import MultiPolygon
```
So we wouldn't have to mock them, and then we wouldn't run afoul of the `isinstance` issue flagged by Riley.

`parse_bounding_polygon_from_wkt` should do the "right thing" for both types of polygons, so I'm fine with just passing through whatever we get from `get_frame_geodataframe`  without checking to see if its `MultiPolygon` or not.

src/opera/test/pge/disp_s1/test_disp_s1_pge.py
```diff
@@ -678,7 +680,6 @@ def test_disp_s1_pge_validate_inputs(self):
                            ionosphere_files + troposphere_files)
         files_to_remove.append(mask_file)
         files_to_remove.append(dem_file)
-        files_to_remove.append(frame_to_burst_file)
```
collinss-jpl, line 681: Since we're using the frame-geometries-simple file, and not the frame-to-burst database, should this line be put back?

collinss-jpl (changes requested):
Looks solid, but I'm not sure we need to take the extra step of translating the anti-meridian polygon before it goes into the ISO XML. See comment below.

src/opera/util/geo_utils.py
kevinmarlis: Whoa, good to know.

RKuttruff (approved):

PR #732, opened 2025-09-17
Update Traceback Regex

### LINKED ISSUE #731: [Bug]: PGE logging fails to capture SAS tracebacks raised from message-less exceptions
### Checked for duplicates

Yes - I've already checked

### Describe the bug

DIST-S1 SAS failures have been noticed to be missing root error tracebacks in HySDS.

This has been traced to errors raised without an associated message, ie:
```python
assert 1 == 2
raise RuntimeError()
```

as opposed to 

```python
assert 1 == 2, 'Mismatch'
dict(foo='bar')['baz']
raise RuntimeError('Something went wrong!!')
```

The root cause of this bug is [the regex](https://github.com/nasa/opera-sds-pge/blob/release/dist-s1/6.0.0-rc.2.0/src/opera/util/run_utils.py#L80-L82) used to capture tracebacks from SAS output logging does not capture these tracebacks (it expects `<exception_name>: <some text>`)

### What did you expect?

I expected any SAS error to have its traceback captured for HySDS

RKuttruff:
Minor patch for regex bug where SAS exceptions without associated message data were not captured.

## Affected Issues
- Closes #731 

## Testing
- Unit tests passing locally

### DIFF
--- a/src/opera/test/data/test_sas_log_with_traceback_no_message.txt
+++ b/src/opera/test/data/test_sas_log_with_traceback_no_message.txt
@@ -1,0 +1,174 @@
+<Example SAS output follows>
+Despeckling batch:   0%|          | 0/1 [00:00<?, ?it/s]
+
+Despeckling:   0%|          | 0/74 [00:00<?, ?it/s][A
+
+Despeckling:   1%|▏         | 1/74 [00:06<07:30,  6.18s/it][A
+
+Despeckling:   4%|▍         | 3/74 [00:06<01:56,  1.64s/it][A
+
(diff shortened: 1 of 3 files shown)


collinss-jpl (changes requested):
To fully close the loop on this bug, we should add an additional test case [here](https://github.com/nasa/opera-sds-pge/blob/main/src/opera/test/util/test_run_utils.py#L202) that reads the `AssertionError` stack trace that surfaced this bug to ensure the fix now parses as we expect.

RKuttruff:
Done

collinss-jpl (approved):

PR #737, opened 2025-09-26
Patch for UTM bbox to lat/lon utility function

### LINKED ISSUE #736: [Bug]: `util.geo_utils.translate_utm_bbox_to_lat_lon` gives incorrect results for some source projections
### Checked for duplicates

Yes - I've already checked

### Describe the bug

This issue was detected by ASF in the ISO XML bounding box for RTC-S1-STATIC product `OPERA_L2_RTC-S1-STATIC_T010-019886-IW2_20140403_S1A_30_v1.0 `. The bounding box from the product was `330420.0, -379770.0,  429540.0, -343080.0` (xmin, ymin, xmax, ymax) in EPSG:3414 projection (WGS 84 / NSIDC Sea Ice Polar Stereographic North). When fed through `util.geo_utils.translate_utm_bbox_to_lat_lon`, this gives an EPSG:4326 bounding box of `(85.35549684400532, 84.92842207070258, -3.975004979772367, 6.385116581294131)` (lat_min, lat_max, lon_min, lon_max). Note the lat_min value is greater than the lat_max value, which is causing errors with ASF on ingest.

The cause of this issue is the assumption that the upper right corner (xmax, ymax) and lower left corner (xmin, ymin) are the max lat/lon and min lat/lon, respectively. For this example product, at its latitude and projection, the upper right corner actually ends up at a lower latitude than the lower left corner.

Image
Image
Image
Image

### What did you expect?

I expected the bbox provided by `util.geo_utils.translate_utm_bbox_to_lat_lon` to be a tuple (lat_min, lat_max, lon_min, lon_max) where lat_min < lat_max and lon_min < lon_max.

### Reproducible steps

```bash
from opera.util.geo_utils import translate_utm_bbox_to_lat_lon
import numpy as np

bounding_box = np.array([330420., -379770., 429540., -343080.])
epsg_code =  3413

translate_utm_bbox_to_lat_lon(bounding_box, epsg_code)
```

### Environment

```bash
opera-dev-pge on commit 090251a0ebdd656a696d444d62609e37a7240590
```

RKuttruff:
## Description
- This PR addresses an issue where the `util.geo_utils.translate_utm_bbox_to_lat_lon` function would return incorrect min/max lat/lon values for certain inputs. This was due to the assumption that the upper right corner == max lat/lon and lower left corner == min lat/lon, which is not always correct for some projections. This PR fixes this issue by instead evaluating all corners of the input bounding box.

## Affected Issues
- Fixes #736 

## Testing
- Added new unit test case for sample product that discovered this bug
- Updated existing test case as it also used a bbox affected by this bug
- Verified tests are passing on dev-pge

### DIFF
--- a/src/opera/util/geo_utils.py
+++ b/src/opera/util/geo_utils.py
@@ -89,8 +89,8 @@ def translate_utm_bbox_to_lat_lon(bbox, epsg_code):
     # Project all corners to EPSG 4326 (lat/lon) then compute mins and maxes.
     # This is necessary since RTC-S1-STATIC projections yield products that are "rotated" w.r.t. the equator when
     # projected into EPSG 4326, and therefore we cannot guarantee which product corner represents the product's
-    # min/max lat/lon. In some cases, this yielded invalid bounding boxes where min >= max lat/lon, which caused
-    # ingest issues downstream at ASF.
+    # min/max lat/lon. In some cases, such assumptions yielded invalid bounding boxes where min >= max lat/lon,
+    # which caused ingest issues downstream at ASF.
     #
     # See: https://github.com/nasa/opera-sds-pge/issues/736 for an example annotated illustration of this issue.
     ulc = (xmin, ymax)
--- a/src/opera/util/geo_utils.py
+++ b/src/opera/util/geo_utils.py
@@ -86,6 +86,13 @@ def translate_utm_bbox_to_lat_lon(bbox, epsg_code):
     elevation = 0
     xmin, ymin, xmax, ymax = bbox
 
+    # Project all corners to EPSG 4326 (lat/lon) then compute mins and maxes.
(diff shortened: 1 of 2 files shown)


src/opera/util/geo_utils.py
```diff
@@ -86,8 +86,20 @@ def translate_utm_bbox_to_lat_lon(bbox, epsg_code):
     elevation = 0
     xmin, ymin, xmax, ymax = bbox
 
-    lat_min, lon_min, _ = transformation.TransformPoint(xmin, ymin, elevation)
-    lat_max, lon_max, _ = transformation.TransformPoint(xmax, ymax, elevation)
+    ulc = (xmin, ymax)
```
collinss-jpl, line 89: Let's include a comment here about the approach being taken and why its needed (correctly handle high-latitude tiles etc..)

collinss-jpl (approved):

PR #798, opened 2026-04-20
Add RFI_FRAMES_COUNT to DSWx-NI measured parameters

RKuttruff:
## Description
- This PR adds the RFI_FRAMES_COUNT to the DSWx-NI measured parameters config.
- Currently, it looks like this field can incorrectly be set to `None`, so we are checking for that and correcting it to `0`.

## Affected Issues
- [OPERA-2412](https://hysds-core.atlassian.net/browse/OPERA-2412)

## Testing
- [x] Updated mock metadata to include field (with "incorrect" None value)
- [x] Ran unit tests on local & dev
- [x] Verified via manual INT test that ISO attribute is correctly populated (+ `None` translated correctly to `0`)
- [x] Jenkins unit tests


kevinmarlis (approved):
LGTM

PR #803, opened 2026-06-09
DSWx-NI PGE CalVal SAS v0.4.1 Integration

RKuttruff:
Integrate latest CalVal SAS v0.4.1.

## Affected Issues
- [OPERA-2529](https://hysds-core.atlassian.net/browse/OPERA-2529)

## Testing
**INT testing must be performed on Intel hardware**

- [x] Manual unit tests on local
- [x] Manual unit tests on dev
- [x] Manual int tests on dev
- [x] Jenkins unit tests
- [x] Jenkins int tests


.ci/scripts/dswx_ni/dswx_comparison.py
kevinmarlis, line 25: Maybe add a comment for why magic numbers and/or what they mean.

kevinmarlis, line 91: Would it be helpful to have the exhaustive list of errors instead of returning just the first instance?

kevinmarlis, line 142: Kind of a minor thing but `flag_all_ok` mutability makes it a little confusing to trace through. Would something like this work?

```
def _get_prefix_str(flag_same):
    return "[OK]" if flag_same else "[FAIL]"

flag_all_ok = True

...

if not flag_same_metadata:
    flag_all_ok = False
    print(prefix + metadata_error_message)

return flag_all_ok
```

kevinmarlis (approved):
Looks good! Just a few minor things that don't necessarily need addressing.

.ci/scripts/dswx_ni/dswx_comparison.py
RKuttruff: This script was copied from the DSWx-S1 comparison script. It used to be a symlink to that but I needed to add the updated metadata skipping and the tolerance override above. I generally want to avoid messing with it

RKuttruff: See my reply to the above comment

RKuttruff: Done

